PR #64, opened 2017-06-13
S3 plugin compatibility

Screenfeed:
The branch `improvement/IM-16-coding-standards` must be merged before.


inc/3rd-party/amazon-s3-and-cloudfront/inc/classes/class-imagify-as3cf.php
```diff
@@ -0,0 +1,280 @@
...
+		return self::$_instance;
+	}
+
+	/**
+	 * The class constructor.
+	 *
+	 * @since  1.6.6
+	 * @author Grégory Viguier
+	 */
+	protected function __construct() {
```
Tabrisrp: Might be better to use a static method to call the hooks outside of the __construct() (see https://carlalexander.ca/designing-class-wordpress-hooks/)

Screenfeed: Indeed I see lots of "don't put your hooks in the constructor" hype stuff lately, on [Tom McFarlin's blog](https://tommcfarlin.com/wordpress-plugin-constructors-hooks/) for example.

Honestly, I don't really get what problems it could solve, especially in this case. But since I'm not especially against it and mainly because it seems better for our future unit tests, I'll go with it.
But, there's one thing I don't like though, the fact that this init method should be static.
Meaning:
- We must hardcode the name of the function to launch this init. That means we'll have to do that: `add_action( 'imagify_loaded', array( 'Imagify_AS3CF', 'init' ), 1 );`, so that means a 3rd party plugin that wants to unhook it, will also hardcode the class name.
- We have a "API function" to get the class instance: `imagify_as3cf()`.
- With the 2 previous points: if someday we feel the need for this class to be extended (to ease 3rd party plugins work for example), we can add a filter inside `imagify_as3cf()` to change the class name. That way we keep compatibility with any `remove_action()`.

Conclusion: ok for the `init()` method, but I think it's better if it's not static:

    `add_action( 'imagify_loaded', array( imagify_as3cf(), 'init' ), 1 );`

What do you think?

Tabrisrp: I agree

Screenfeed: OK thanks, I'll go with it then.

Screenfeed:
@Tabrisrp Done :)

PR #86, opened 2017-07-10
Fixes #76: add support for S3 Pro

### LINKED ISSUE #76: Compatibility with the plugin WP Offload S3 PRO
Now we have compatibility with the Free version, add the Pro.

https://secure.helpscout.net/conversation/388794094/43520/?folderId=676700

Screenfeed:
It also fixes several bugs related to files being located only on S3 (not on the local server).

### DIFF
--- a/inc/classes/abstracts/class-imagify-abstract-attachment.php
+++ b/inc/classes/abstracts/class-imagify-abstract-attachment.php
@@ -23,7 +23,7 @@ class Imagify_Abstract_Attachment {
 	 * @var int
 	 * @access public
 	 */
-	public $id;
+	public $id = 0;
 
 	/**
 	 * The constructor.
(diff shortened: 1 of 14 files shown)


inc/classes/abstracts/class-imagify-abstract-attachment.php
```diff
@@ -33,7 +33,21 @@ class Imagify_Abstract_Attachment {
...
+				$this->id = $id->ID;
+			} elseif ( is_numeric( $id ) ) {
+				$this->id = $id;
+			}
+		} elseif ( $post && is_a( $post, 'WP_Post' ) ) {
+			$this->id = $post->ID;
+		}
+
+		$this->id = absint( $this->id );
+	}
```
Tabrisrp: Should an abstract class have any code in it? Isn't it supposed to lay ground for the base structure only?

Screenfeed: Well, if you look at the whole class, more than half of the methods already contain code, so... >_>

Tabrisrp: right it's abstract not interface, my bad

PR #120, opened 2017-08-28
Fixes #116

### LINKED ISSUE #116: Installing via WordPress packagist not working
For some reason, installing this plugin with composer from WordPress Packagist (https://wpackagist.org/) did broke last Saturday. In wpackagist, the plugin seems to be inactive.

Message from composer is following

> Your requirements could not be resolved to an installable set of packages.
> 
>   Problem 1
>     - The requested package wpackagist-plugin/imagify could not be found in any version, there may be a typo in the package name.
> 
> Potential causes:
>  - A typo in the package name
>  - The package is not available in a stable-enough version according to your minimum-stability setting
>    see <https://getcomposer.org/doc/04-schema.md#minimum-stability> for more details.

This is kinda major, because running `composer update` brokes when Imagify is required.

Screenfeed:
Missing fields in `composer.json`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,19 @@
 {
   "name": "wpmedia/imagify-plugin",
+  "type": "wordpress-plugin",
+  "homepage": "https://imagify.io",
+  "license": "GPL-2.0",
+  "authors": [
+    {
+      "name": "WP Media",
+      "homepage": "https://wp-media.me/"
+    },
+    {
+      "name": "Grégory Viguier",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,5 +1,19 @@
 {
   "name": "wpmedia/imagify-plugin",
```
Tabrisrp: should be wp-media no?

Screenfeed: Indeed, better.

PR #135, opened 2017-09-29
Don't resize or stripe EXIF for NGG images

mweichert:
As discussed, NGG has it's own resizing logic and to avoid integrity problems, we'd like to leave NGG responsible for resizing images. 

A lot of our userbase are photographers, and therefore never want EXIF meta data stripped.


Screenfeed:
I was expecting only an issue to be open, I've got the full PR. Thanks ^^
I must adapt the code to our standards but that's a great help.

Screenfeed:
Fixed in this commit: https://github.com/wp-media/imagify-plugin/pull/155/commits/4d53f3a4cdab5bf4cff35fb27771e696330f991c.
Thanks.

PR #146, opened 2017-10-12
Fixes #142: Statistics Layout could be broken

### LINKED ISSUE #142: Statistics Layout could be broken
image001

From https://app.intercom.io/a/apps/cd6nxj3z/respond/inbox/38607/conversations/11962696684

Screenfeed:
Fixes the width of the "Optimized size" in the bulk optimization page.


Screenfeed:
Yes that's intended, I never minify CSS and JS files in a PR, it prevents conflicts with other branches that would also modify those files.

PR #156, opened 2017-10-19
Fixes #139

### LINKED ISSUE #139: That's the number of original images you optimized with Imagify
In Bulk Optimization under "Statistics",
change "that's the number of images you optimized with Imagify"

to

"that's the number of original images you optimized with Imagify"

Screenfeed:


inc/admin/ui/bulk.php
```diff
@@ -138,7 +138,7 @@ function _imagify_display_bulk_page() {
 								<?php
 								printf(
 									/* translators: you can use %s to include a line break. */
-									__( 'that\'s the number of images you optimized with Imagify', 'imagify' ),
+									__( 'that\'s the number of original images you optimized with Imagify', 'imagify' ),
```
Tabrisrp: Not directly related to the fix, but in WP Rocket it was decided to not use abbreviation for things like that's, you're, etc…notably to prevent the backslash use.

Screenfeed: I also avoid them when possible.

PR #159, opened 2017-10-20
Improvement/more objects

Screenfeed:
BASED ON BRANCH improvement/get-attachment!

### DIFF
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -82,9 +82,7 @@ function _imagify_new_upgrade( $imagify_version, $current_version ) {
 
 		if ( $query->posts ) {
 			foreach ( (array) $query->posts as $id ) {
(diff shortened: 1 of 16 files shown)


inc/admin/upgrader.php
```diff
@@ -95,8 +91,10 @@ function _imagify_new_upgrade( $imagify_version, $current_version ) {
 		}
 
 		// Auto-activate the Admin Bar option.
-		$options                   = get_site_option( IMAGIFY_SETTINGS_SLUG );
+		$options = get_site_option( IMAGIFY_SETTINGS_SLUG );
+		$options = is_array( $options ) ? $options : array();
```
Tabrisrp: `$options = (array) $options;` directly?

Screenfeed: No, because if `$options` is `false`, doing `$options = (array) $options;` will result in `array( false )`.

PR #173, opened 2017-11-01
Introduces class `Imagify_DB`

Screenfeed:
This is the last one for this version.

### DIFF
--- a/inc/3rd-party/nextgen-gallery/inc/classes/class-imagify-ngg-db.php
+++ b/inc/3rd-party/nextgen-gallery/inc/classes/class-imagify-ngg-db.php
@@ -109,8 +109,6 @@ class Imagify_NGG_DB extends Imagify_Abstract_DB {
 	 * @author Jonathan Buttigieg
 	 */
 	public function create_table() {
-		global $wpdb;
-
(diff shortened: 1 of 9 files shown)


inc/3rd-party/nextgen-gallery/inc/classes/class-imagify-ngg-db.php
```diff
@@ -29,9 +29,9 @@ class Imagify_NGG_DB extends Imagify_Abstract_DB {
...
 	 *
-	 * @access  public
-	 * @since   1.5
-	 * @author  Jonathan Buttigieg
+	 * @since  1.5
+	 * @access protected
+	 * @author Jonathan Buttigieg
 	 */
 	protected function __construct() {
 		global $wpdb;
```
Tabrisrp: can remove `$wpdb` since it's not used here anymore

Screenfeed: It is used 3 times :|
But not in `create_table()` indeed.

Tabrisrp: comment didn't go where I wanted to, it was for `create_table()` specifically

PR #182, opened 2017-11-16
Large library transient

Screenfeed:
Thinking it wasn't used, the first idea was to remove the `imagify_large_library` transient.
But in fact it is used, it's just a bit messed up.

TL;DR: consistency, small fixes, new helper function.


Screenfeed:
Don't you forget about me...

Don't you forget about me

PR #196, opened 2018-01-15
Custom folders part 2: database

Screenfeed:
Let's pray I succeeded in fixing Github apocalypse.


Screenfeed:
@Tabrisrp I suggest you don't look the whole "Files changed" tab, but go commit by commit instead.
Github atomized everything and I tried to recover code in the final commit. Honestly I don't even know how this will end, maybe some file or code will be missing, this is my worst nightmare.

Tabrisrp:
now you 🙏 I don't delete the branch just in case

Screenfeed:
We'll keep this one just in case yes, but a deleted branch can be reverted anyway.

PR #199, opened 2018-01-18
Custom folders part 3: various

Screenfeed:


Screenfeed:
For this one you can ignore the 3 last commits: deprecated.php and all notice templates.

PR #201, opened 2018-01-23
Custom folders part 5

Screenfeed:
Introduced classes `Imagify_Files_Scan`, `Imagify_Files_Iterator`, and `Imagify_Files_Recursive_Iterator`, allowing to scan (recursively or not) for files to optimize.


Screenfeed:
As usual, no need to look into the "Merge" commit.

inc/classes/class-imagify-files-scan.php
```diff
@@ -0,0 +1,500 @@
...
+	public function get_files_from_folder( $folder ) {
+		// Formate and validate the folder path.
+		if ( ! is_string( $folder ) || '' === $folder || '/' === $folder || '\\' === $folder ) {
+			return new WP_Error( 'invalid_folder', __( 'Invalid folder.', 'imagify' ) );
+		}
+
+		$folder = realpath( $folder );
+
+		if ( ! $folder ) {
+			return new WP_Error( 'folder_not_exists', __( 'This folder doesn\'t exist.', 'imagify' ) );
```
Tabrisrp: for translators, what we did on WP Rocket is to use the full negative form instead of the contraction (does not instead of doesn't), this prevent having to escape the quote too

Screenfeed: I also try to, but I forget sometimes, and I admit I prefer the contraction sometimes, like here: in most cases I prefer "doesn't" over "does not", but I prefer "is not" over "isn't".
But you're right, I should use "does not". Moreover, we have "This file is not a folder" and "This folder is not allowed" few lines later, so it makes sense.

PR #215, opened 2018-02-09
Custom folders part 11

Screenfeed:
CRON to sync files and DB.


inc/functions/custom-folders.php
```diff
@@ -597,3 +619,173 @@ function imagify_insert_custom_file( $args = array() ) {
...
+		 */
+		if ( $backup_path ) {
+			// Delete the backup file.
+			$filesystem->delete( $backup_path );
+		}
+
+		// Remove the entry from the database.
+		$file->delete_row();
+
+		return new WP_Error( 'no-file', __( 'The file was missing or its path couldn\'t be retrieved from the database. The entry has been deleted from the database.', 'imagify' ) );
```
Tabrisrp: I'm the annoying guy about contraction :D

inc/functions/custom-folders.php
```diff
@@ -597,3 +619,173 @@ function imagify_insert_custom_file( $args = array() ) {
...
+		}
+	}
+
+	// Hash + modified.
+	$current_hash = md5_file( $file_path );
+
+	if ( ! $old_data['hash'] ) {
+		$new_data['modified'] = 0;
+	} else {
+		$new_data['modified'] = (int) ! hash_equals( $old_data['hash'], $current_hash );
```
Tabrisrp: you have a compat for hash_equals? (since it's PHP 5.6)

Screenfeed: You're right, it was missing. The polyfill appeared in WP only in version 3.9.2.

Screenfeed:
Both fixed :)

PR #230, opened 2018-03-06
Fixes #197

### LINKED ISSUE #197: Improve tests for AS3
I found a case where `$as3cf` was `null`, triggering a fatal error when trying to access it as an object.
[Source](https://app.intercom.io/a/apps/cd6nxj3z/respond/inbox/1295841/conversations/14086652972).

Screenfeed:
Following PR #204, fixing #197:
- Keep compat' with old versions of the plugin WP Offload S3 that require the plugin Amazon Web Services.
- Keep compat' with new versions that do not require the plugin Amazon Web Services.


Screenfeed:
Sorry @Tabrisrp, this is required if we don't want to break everything with the new version of WP Offload S3 ;)

PR #237, opened 2018-03-16
Improvement: filesystem, fixes #235

### LINKED ISSUE #235: Don't use `wp_mkdir_p()`
On a server with IIS: if the site's path begins with `\\`, `wp_mkdir_p()` will fail.
The path must be normalized first, but then:
```php
$target = str_replace( '//', '/', $target );
```
The path will now start with a single `/`, and `file_exists()` will fail.

This is used to create the backup folders.
[Source](https://app.intercom.io/a/apps/cd6nxj3z/inbox/inbox/conversation/15090717723).

Screenfeed:
Introduced class Imagify_Filesystem.
Among other things, it should also fixes #235.


Screenfeed:
You're right for both.
- For the property I was worry about memory and duplicating the same same data over and over, but it doesn't seem to be an issue since (as far as I know) the object is not duplicated.
- Right, I should deprecate them, they are not needed anymore.

PR #250, opened 2018-04-11
Fixes #236: OPcache reset

### LINKED ISSUE #236: Clear OPcache on update
If OPcache is not cleared when Imagify is updated, it triggers a fatal error if the new version removed files.
It's not related to Imagify itself, it can happen on any core, plugin, or theme update.

Screenfeed:
Change the hook used to reset opcache.
This one is safer because it is triggered after the maintenance mode is disabled. Also, this is the same hook that WordPress will use when this functionnality is released in core.

### DIFF
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -271,34 +271,43 @@ function _imagify_new_upgrade( $network_version, $site_version ) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Screenfeed:
Sorry Remy for this one. I changed the hook used for that, this one seems safer.

PR #262, opened 2018-06-21
Fixes #258: support for content folders outside WP folder

### LINKED ISSUE #258: Custom folders architecture not supported anymore
The following doesn't work anymore:
```
/wp/wp-admin/
/wp/wp-includes/
/wp-content-custom/
```
The issue is that the "wp-content" folder is outside `ABSPATH`, while we validate all paths against it (`ABSPATH`'s value is `/wp/` here).

[Source](https://secure.helpscout.net/conversation/581160833/68260?folderId=1724287).

Screenfeed:
Basically, `ABSPATH` is not the path site's root folder.

### DIFF
--- a/inc/classes/class-imagify-filesystem.php
+++ b/inc/classes/class-imagify-filesystem.php
@@ -17,7 +17,7 @@ class Imagify_Filesystem extends WP_Filesystem_Direct {
 	 *
 	 * @var string
 	 */
-	const VERSION = '1.1';
+	const VERSION = '1.2';
 
 	/**
 	 * Delimiter used for regex patterns.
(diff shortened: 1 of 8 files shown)


inc/classes/class-imagify-filesystem.php
```diff
@@ -713,7 +722,109 @@ public function is_root( $path ) {
...
+	 * Get the path to the site's root.
+	 *
+	 * @since  1.8.1
+	 * @access public
+	 * @see    get_home_path()
+	 * @author Grégory Viguier
+	 *
+	 * @return string
+	 */
+	public function get_site_root() {
```
Tabrisrp: you didn't have any issue with different setups and this function?

When I tried to copy get_home_path() in WP Rocket, it didn't return the values I expected

Screenfeed: So far it worked as expected with everything I tested.

Screenfeed: The only problem with `get_home_path()` is on multisite (surprise!), especially when not on the main site. That's why I use this modified version of it, instead of the real one.

PR #293, opened 2018-09-04
Fixes #280: Display a loader when interacting with the custom files tree

### LINKED ISSUE #280: Display a loader when interacting with the custom files tree
When browsing the custom files tree, a loader icon should be displayed after clicking the "+" icon (to fetch files inside the clicked folder): at this moment an ajax request is done and it can take 1 or 2 seconds.

Screenfeed:


Tabrisrp:
Conflict to fix to be able to merge

PR #336, opened 2019-04-10
Webp: Use new functions and classes

Screenfeed:
Also rename few things, like hooks and ajax callbacks.


inc/classes/class-imagify-admin-ajax-post.php
```diff
@@ -375,132 +438,80 @@ public function imagify_bulk_upload_callback() {
...
+		imagify_check_nonce( 'imagify_reoptimize_file' );
 		imagify_check_user_capacity( 'optimize-file' );
 
-		$file = $this->get_file_to_optimize( 'imagify_optimize_file' );
-
-		// Optimize it.
-		$result = $file->optimize();
+		$media_id = $this->get_media_id( 'GET', 'id' );
+		$level    = $this->get_optimization_level( 'GET', 'level' );
+		$process  = imagify_get_optimization_process( $media_id, 'custom-folders' );
```
Tabrisrp: You set `$process`, but don't use it?

Screenfeed: This will change in a future branch, `$process` won't be set.

PR #340, opened 2019-04-17
Webp: display webp images on front-end

Screenfeed:


Screenfeed:
Thanks, eagle eye, I've added those 2 to the things to fix in the final branch.

PR #361, opened 2019-05-03
Generate webp versions from the settings page

Screenfeed:
This adds a button in the settings page to generate webp versions for the images that are already optimized by Imagify.
This also removes the action selector from the bulk optimization page.


inc/classes/class-imagify-admin-ajax-post.php
```diff
@@ -521,7 +477,18 @@ public function imagify_manual_reoptimize_callback() {
 			imagify_die();
 		}
 
-		$this->reoptimize_media( $media_id, $context, $this->get_optimization_level() );
+		$rsult = $this->reoptimize_media( $media_id, $context, $this->get_optimization_level() );
```
Tabrisrp:
```suggestion
		$result = $this->reoptimize_media( $media_id, $context, $this->get_optimization_level() );
```

PR #362, opened 2019-05-03
Various last changes before 1.9

Screenfeed:


inc/functions/common.php
```diff
@@ -36,6 +36,20 @@ function imagify_get_context_names() {
...
+ * Sanitize an optimization context.
+ *
+ * @since  1.6.11
+ * @author Grégory Viguier
+ *
+ * @param  string $context The context.
+ * @return string
+ */
+function imagify_sanitize_context( $context ) {
+	$context = preg_replace( '/[^a-z0-9_-]/i', '', $context );
```
Tabrisrp: You could use `sanitize_key()` here to simplify, you would not have to use the `strtolower()` on the following line

Screenfeed: You’re right, `sanitize_key()` does exactly that. I’m not sure I want to add a filter here though (the function has a `sanitize_key` filter).

Screenfeed: OK, done.

PR #363, opened 2019-05-03
Update class-imagify-abstract-attachment.php

alimuzzaman:
Allow 3rd party plugin to filter hs_backup().
Backup maybe not available in server, only accessible using CDN.

Our **WP Stateless** plugin remove attachment files from server (in **Stateless** Mode), after uploading to GCS (Google Cloud Storage). 
So we need a way to tell Imagify that backup exists in GCS, even though not exist in server.
When **Imagify** tries to restore we use `before_imagify_restore_attachment` to restore backup image if not exists to server from GCS.

### DIFF
--- a/inc/classes/class-imagify-abstract-attachment.php
+++ b/inc/classes/class-imagify-abstract-attachment.php
@@ -252,8 +252,9 @@ abstract class Imagify_Abstract_Attachment extends Imagify_Abstract_Attachment_D
 		}
 
 		$backup_path = $this->get_raw_backup_path();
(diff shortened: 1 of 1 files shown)


Screenfeed:
Hi.

The version 1.9 has just been released and the optimization process has been totally rewritten. So this part is outdated, the class has been deprecated.
In the new class [there is still this method](https://github.com/wp-media/imagify-plugin/blob/1f27c546b7ab2b926adc6c89db8b8892df6ff5b4/classes/Media/AbstractMedia.php#L254-L275), and its whole purpose is to test the file existence. So the chances that we allow such filter here are near 0.

PR #402, opened 2019-07-05
Closes #392: Don't save WebP when it's bigger than jpg

### LINKED ISSUE #392: Don't save WebP when it's bigger than jpg
When a WebP is created, we don't check if it's bigger than the jpg.
I think we should check the size and if it's bigger don't save it (only for the new images though)

Screenfeed:


Screenfeed:
Old on for this one. Drama happening.

Screenfeed:
OK, drama ended, review can resume.

Screenfeed:
o/

PR #435, opened 2019-10-14
Compatibility with WP 5.3: auto-resize

Screenfeed:
WP 5.3 will resize images on upload. This PR adds compatibility with this new process.
Basically, a new concept is introduced: the full size file is not the original file anymore. So, we need to provide methods handling both files, without breaking other contexts (custom files, NGG) where they are the same file.
This also deals with change in the name of the full size file: `my-image.jpg` will become `my-image-resized-2560.jpg` after WP's resizing (it may also become `my-image-rotated.jpg` if rotated without resizing).
Compatibility with AS3 and NGG have been updated.

### DIFF
--- a/classes/CDN/PushCDNInterface.php
+++ b/classes/CDN/PushCDNInterface.php
@@ -89,7 +89,7 @@ interface PushCDNInterface {
 	 * @access public
 	 * @author Grégory Viguier
 	 *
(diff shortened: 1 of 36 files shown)


Screenfeed:
dont-you-forget-about-me

Screenfeed:
Right, we got a version 1.9.7 released in the mean time. This is fixed with `1.9.8`.

PR #479, opened 2020-01-20
Closes #408: don't say the user is over quota when cannot fetch data

### LINKED ISSUE #408: Deal with cases where user data can't be fetched
During a bulk optimization, a "out of quota" message can be displayed while the user still has quota left.
So far the assumption is that it could be triggered because we can't fetch the user data temporarily, because of a timeout or any other reason.

Screenfeed:
We do mainly two things:
- If an error is returned when fetching user account data, `is_over_quota()` will return `false` now.
- When testing for requirements (cURL is available, image editor is available, requests to Imagify’s API are not blocked internally, API is up, API key is valid, user account is over quota), the "is over quota" test will only test the quota, not taking other tests into account (blocked, API up, key valid).

This has been tested by forcing `Imagify->get_user()` to return a `WP_Error` object.

Closes #408.


inc/classes/class-imagify-user.php
```diff
@@ -152,6 +170,10 @@ public function __construct() {
 	public function get_percent_consumed_quota() {
 		static $done = false;
 
+		if ( $this->get_error() ) {
+			return 0;
+		}
+
```
hellofromtonya: Does having an error mean that the process is not done, i.e. `$done -= false;`?

classes/Imagifybeat/Actions.php
```diff
@@ -66,18 +66,13 @@ public function add_requirements_to_response( $response, $data ) {
...
 			'curl_missing'          => ! \Imagify_Requirements::supports_curl(),
 			'editor_missing'        => ! \Imagify_Requirements::supports_image_editor(),
-			'external_http_blocked' => $is_blocked,
-			'api_down'              => $is_blocked || ! $is_api_up,
-			'key_is_valid'          => ! $is_blocked && $is_api_up && $is_key_valid,
-			'is_over_quota'         => ! $is_blocked && $is_api_up && $is_key_valid && $is_over_quota,
+			'external_http_blocked' => \Imagify_Requirements::is_imagify_blocked(),
+			'api_down'              => ! \Imagify_Requirements::is_api_up(),
+			'key_is_valid'          => \Imagify_Requirements::is_api_key_valid(),
+			'is_over_quota'         => \Imagify_Requirements::is_over_quota(),
```
hellofromtonya: Instead of using `\` to indicate the class is in the global namespace, import the class at the top of the file, i.e. after the `namespace` declaration using `use Imagify_Requirements`. 

Why?

Easier to read. The backslashes can be distracting.

inc/classes/class-imagify-user.php
Screenfeed: `$done` is only used to prevent hooks to be triggered several times for nothing, so it has no meaning for the process itself.

classes/Imagifybeat/Actions.php
```diff
@@ -66,18 +68,13 @@ public function add_requirements_to_response( $response, $data ) {
...
-			'external_http_blocked' => $is_blocked,
-			'api_down'              => $is_blocked || ! $is_api_up,
-			'key_is_valid'          => ! $is_blocked && $is_api_up && $is_key_valid,
-			'is_over_quota'         => ! $is_blocked && $is_api_up && $is_key_valid && $is_over_quota,
+			'curl_missing'          => ! Requirements::supports_curl(),
+			'editor_missing'        => ! Requirements::supports_image_editor(),
+			'external_http_blocked' => Requirements::is_imagify_blocked(),
+			'api_down'              => ! Requirements::is_api_up(),
+			'key_is_valid'          => Requirements::is_api_key_valid(),
+			'is_over_quota'         => Requirements::is_over_quota(),
```
hellofromtonya: Let's be consistent. [Here](https://github.com/wp-media/imagify-plugin/pull/479/files#diff-c2c4f325ab724701d7fae667b2701ffaR83) the code uses the full class name instead of an alias. Typically we only want to use an alias _when_ there's a conflict with class names.

I'd suggest removing the `as Requirements` alias and using the full class name on lines 72-77.

Screenfeed: We also use alias to shorten our class names and make the code easier to read, like using `Foobar` instead of `\SubVeryLong1\SubVeryLong2\SubVeryLong3\SubVeryLong4\Foobar`.

Screenfeed:
Added unit and integration tests.

tests/Integration/inc/classes/ImagifyUser/TestGetPercentConsumedQuota.php
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Test Imagify_User->get_percent_consumed_quota() should return 0 when couldn’t fetch user account data.
+	 */
+	public function testShouldReturnZeroWhenCouldNotFetchUserData() {
+		$wp_error = new WP_Error( 'error_id', 'Error Message' );
+
+		Functions\when( 'get_imagify_user' )->justReturn( $wp_error );
+		Functions\expect( 'imagify_round_half_five' )->never();
+
+		$this->assertSame( ( new Imagify_User() )->get_percent_consumed_quota(), 0 );
```
hellofromtonya: Instead mocking the functions, let them run. How? Before instantiating the user, set up the conditions for the API not to find and return the user's data. 

Think about what would cause this state, such as the user doesn't have an account. Once you know what that is, then look for a way to hook into `Imagify::get_user()` to trigger that state when calling the API, e.g. pass it an invalid user credentials.

Another tip: Move the `static $done` to be a static property on the class. Then use  `$this-set_reflective_property()` to reset it after each  test.

```php
// Set up the "couldn't fetch user account data" condition here.

$user = new Imagify_User();
$this->assertInstanceOf( 'WP_Error', $user->get_error() );
$this->assertSame( 0, $user-> get_percent_consumed_quota() );
```

tests/Integration/inc/classes/ImagifyUser/TestGetPercentConsumedQuota.php
```diff
@@ -0,0 +1,60 @@
...
+		Imagify_Data::get_instance()->set( 'previous_quota_percent', 90.0 ); // Previous quota was 90%.
+
+		$this->assertSame( ( new Imagify_User() )->get_percent_consumed_quota(), 90.0 );
+
+		$userData->consumed_current_month_quota = 500; // Current consumed quota 50%.
+
+		Functions\when( 'get_imagify_user' )->justReturn( $userData );
+
+		$this->assertSame( ( new Imagify_User() )->get_percent_consumed_quota(), 50.0 );
+	}
```
hellofromtonya: Same here, except this time the API should return the user's account data. Once received, you can manipulate the consumed quota to fix your test.

tests/Integration/inc/classes/ImagifyUser/TestIsOverQuota.php
```diff
@@ -0,0 +1,98 @@
+<?php
```
hellofromtonya: Rename the file to `isOverQuota.php`.

tests/Unit/inc/classes/ImagifyUser/TestGetError.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
hellofromtonya: Rename file to `getError.php`

tests/Unit/TestCase.php
```diff
@@ -61,5 +61,10 @@ protected function mockCommonWpFunctions() {
 		foreach ( $functions as $function ) {
 			Monkey\Functions\when( $function )->echoArg();
 		}
+
+		Monkey\Functions\when( 'is_wp_error' )
+			->alias( function() {
+				return $thing instanceof WP_Error;
```
hellofromtonya: `WP_Error` is not loaded into memory in a unit test, as WordPress is not running. You'll need a mock or fixture here.

tests/Unit/inc/classes/ImagifyUser/TestGetPercentConsumedQuota.php
```diff
@@ -0,0 +1,103 @@
+<?php
```
hellofromtonya: Rename the file to `getPercentConsumedQuota.php`.

tests/Unit/inc/classes/ImagifyUser/TestGetPercentConsumedQuota.php
```diff
@@ -0,0 +1,103 @@
...
+					}
+					if ( $decimal <= 0.99 ) {
+						return floatval( $number[0] ) + 1;
+					}
+					return 1;
+				}
+
+				return floatval( $number );
+			}
+		);
```
hellofromtonya: Instead of all of this code, `expect` this function `with` a known amount and then return what is expected. That avoids errors in the code here. For example:

```php
Functions\expect( 'imagify_round_half_five' )
	->once()
	->with( 0 ) // extra quota consumed
	->andReturn( 0 );
```

^ Adjust the numbers for the test.

tests/Unit/inc/classes/ImagifyUser/TestIsOverQuota.php
```diff
@@ -0,0 +1,140 @@
+<?php
```
hellofromtonya: Rename file to `isOverQuota.php`.

tests/Unit/inc/classes/ImagifyUser/TestIsOverQuota.php
```diff
@@ -0,0 +1,140 @@
...
+					}
+					if ( $decimal <= 0.99 ) {
+						return floatval( $number[0] ) + 1;
+					}
+					return 1;
+				}
+
+				return floatval( $number );
+			}
+		);
```
hellofromtonya: See comment above. Use BrainMonkey to return the value you expect from this test.

tests/Unit/inc/classes/ImagifyUser/TestIsOverQuota.php
```diff
@@ -0,0 +1,140 @@
...
+
+		$imagify_data_mock = Mockery::mock( Imagify_Data::class );
+		// Change the Imagify_Data::$_instance to the mock.
+		$ref = $this->get_reflective_property( '_instance', Imagify_Data::class );
+
+		$ref->setValue( $imagify_data_mock, $imagify_data_mock );
```
hellofromtonya: This code could be moved to the trait for reuse throughout the tests.

```php
$this->set_singleton_instance( Imagify_Data::class, $imagify_data_mock );
```

And the method on the Trait:

```php

/**
 * Set the singleton's `$_instance` property to the given instance.
 *
 * @param string $class_name Name of the target class.
 * @param mixed  $instance   Instance of the target object.
 *
 * @throws \ReflectionException Throws an exception if property does not exist.
 */
protected function set_singleton_instance( $class, $instance ) {
	$ref = $this->get_reflective_property( '_instance', $class );
	$ref->setValue( $instance, $instance );
}
```

tests/Unit/inc/classes/ImagifyUser/TestGetPercentConsumedQuota.php
```diff
@@ -0,0 +1,103 @@
...
+
+		$imagify_data_mock = Mockery::mock( Imagify_Data::class );
+		// Change the Imagify_Data::$_instance to the mock.
+		$ref = $this->get_reflective_property( '_instance', Imagify_Data::class );
+
+		$ref->setValue( $imagify_data_mock, $imagify_data_mock );
```
hellofromtonya: This code could be moved to the trait for reuse throughout the tests.

```php
$this->set_singleton_instance( Imagify_Data::class, $imagify_data_mock );
```

And the method on the Trait:

```php

/**
 * Set the singleton's `$_instance` property to the given instance.
 *
 * @param string $class_name Name of the target class.
 * @param mixed  $instance   Instance of the target object.
 *
 * @throws \ReflectionException Throws an exception if property does not exist.
 */
protected function set_singleton_instance( $class, $instance ) {
	$ref = $this->get_reflective_property( '_instance', $class );
	$ref->setValue( $instance, $instance );
}
```

tests/Unit/TestCase.php
Screenfeed: Isn't what I'm doing here?

Screenfeed:
@hellofromtonya Ready for another run.

hellofromtonya:
@Screenfeed Shouldn't this PR be merged into `branch-1.9.9`?

PR #480, opened 2020-01-20
Improved opcache reset

Screenfeed:
Based on an issue found in WP Rocket, use `ini_get( 'opcache.enable' )` to tell if opcache can be reset.


inc/admin/upgrader.php
```diff
@@ -357,6 +357,13 @@ function imagify_reset_opcache() {
 			return;
 		}
 
+		$opcache_enabled = ini_get( 'opcache.enable' );
```
crystinutzaa: Add filter_var() and simplify the bailout condition:
```suggestion
		$opcache_enabled = filter_var( ini_get( 'opcache.enable' ), FILTER_VALIDATE_BOOLEAN );
		
		if ( ! $opcache_enabled ) {
			$can_reset = false;
			return;
		}
```

Screenfeed: Nice, I didn't know this filter could be used that way.

PR #481, opened 2020-01-20
Prevent the "chunky upload" error

### LINKED ISSUE #465: Chunky upload is not supported by HTTP 1.0
Some users get the following error when using Imagify:

`Our server returned an error (0 - Chunky upload is not supported by HTTP 1.0).`

At this moment, we solve this with a helper plugin, however, it looks like we have increased tickets of this kind (I would say at least 2-3 per week).
[source](https://secure.helpscout.net/conversation/1033678461/136876?folderId=1724287).

Screenfeed:
When sending a file for optimization, use HTTP/1.1 by default: the filter `imagify_curl_http_version_1_0` now defaults to `false`.

Fixes #465.

### DIFF
--- a/inc/classes/class-imagify.php
+++ b/inc/classes/class-imagify.php
@@ -526,11 +526,12 @@ class Imagify {
 			 * Tell which http version to use with cURL during image optimization.
 			 *
 			 * @since  1.8.4.1
+			 * @since  1.9.9 Default value is `false`.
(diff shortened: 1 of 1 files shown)


Screenfeed:
@hellofromtonya I won't be able to add tests for this one since we still don't know what triggers this error.

Issue #423, opened 2019-08-14
WebP rewrite rules from NGINX issue


markonikolic985:
It appears to be a problem with NGINX rewrite rules for WebP display.

From a user:

> I got the following error when running nginx -t : 

`- invalid parameter "final" in /var/www/go2roues.com/htdocs/conf/imagify.conf:12 -`

Current recommended NGINX rules by us:

```
location ~* ^(/.+)\.(jpg|jpeg|jpe|png|gif)$ {
    add_header Vary Accept;

    if ($http_accept ~* "webp"){
        set $imwebp A;
    }
    if (-f $request_filename.webp) {
        set $imwebp  "${imwebp}B";
    }
    if ($imwebp = AB) {
        rewrite ^(.*) $1.webp final;
    }
}
```

Meanwhile, suggesting to the user to remove `final` and see if that helps. The ticket can be found here - https://app.intercom.io/a/apps/cd6nxj3z/inbox/inbox/unassigned/conversations/23239228924

aerd2:
Hello, all you have to do is to replace word "final" with "last" and it should work as intended.

iCaspar:
Resolved on #488.

PR #488, opened 2020-04-20
Fix NGINX Rewrite Rules

markonikolic985:
As we have confirmation from one user that removing "final" worked for them, sending this Pull request.

Issue: https://github.com/wp-media/imagify-plugin/issues/423

Ticket: https://secure.helpscout.net/conversation/1138088855/158359?folderId=676702

### DIFF
--- a/classes/Webp/RewriteRules/Nginx.php
+++ b/classes/Webp/RewriteRules/Nginx.php
@@ -45,7 +45,7 @@ location ~* ^(' . $home_root . '.+)\.(' . $extensions . ')$ {
 		set $imwebp  "${imwebp}B";
 	}
 	if ($imwebp = AB) {
-		rewrite ^(.*) $1.webp final;
+		rewrite ^(.*) $1.webp;
 	}
 }' );
 	}


AnthoRdev:
What's happened with this issue? Is that the correct way to solve the problem, removing the "final" parameter?

GeekPress:
@markonikolic985 Can you please explain to @Davidacu how to reproduce the issue, and how to check it's fixed with the new code?

markonikolic985:
@Davidacu 

**How to reproduce the issue**

You need NGINX server with SSH access.

- If possible, connect via SSH and keep the terminal opened

- Activate "Display WebP" using rewrite rules on NGINX server

-  You should already have error, if not - check nginx.config (NGINX configuration file) with `nginx -t`

The error should look like:

`invalid parameter "final" in /nginx_rules_custom.conf:15`

**How to fix the issue**

- Remove `final` from this line: `rewrite ^(.*) $1.webp final;`

- Restart NGINX `systemctl reload nginx`

- Check the syntax again with `nginx -t`

It should be working fine.

iCaspar:
@wp-media/qa Ready for QA.

Davidacu (approved):
- I was able to reproduce the error in 1.9.10
- I can confirm that the `/conf/imagify.conf` file generated does not have **final**
- I can confirm that webp images are served when "rewrite rules" are set
tested in https://spinupwp.wp-rocket.me/sample-page/

hellofromtonya:
@Davidacu Is QA done on this one? If yes, please approve the PR.

Issue #493, opened 2020-05-06
Compatibility with plugin "Enable Media Replace" broken with WP 5.3


Screenfeed:
In its version 3.3.8, the plugin Enable Media Replace changed some things to adapt to WordPress 5.3’s new upload process.
Since this version, our compatibility is broken. Currently, Imagify is hooking `'emr_unfiltered_get_attached_file'` to initiate everything:
```php
add_filter( 'emr_unfiltered_get_attached_file', [ \Imagify\ThirdParty\EnableMediaReplace\Main::get_instance(), 'init' ] );
```
But now in `\EnableMediaReplace\Replacer::__construct()`:
```php
      if (function_exists('wp_get_original_image_path')) // WP 5.3+
      {
          $source_file = wp_get_original_image_path($post_id);
          if ($source_file === false) // if it's not an image, returns false, use the old way.
            $source_file = trim(get_attached_file($post_id, apply_filters( 'emr_unfiltered_get_attached_file', true )));
      }
      else
        $source_file = trim(get_attached_file($post_id, apply_filters( 'emr_unfiltered_get_attached_file', true )));
```
If `wp_get_original_image_path()` doesn’t return `false`, the hook `'emr_unfiltered_get_attached_file'` is not triggered anymore.

This results in the images being improperly optimized, and the old backup file + old webp images not being replaced.
[Plugin](https://wordpress.org/plugins/enable-media-replace/).

Screenfeed:
**Scope a solution** ✅ 
Hook `'wp_handle_replace'` instead of `'emr_unfiltered_get_attached_file'`. The only downside is that this hook is located outside the class `EnableMediaReplace\Replacer`. The perfect solution would be to have a hook available at the beginning of `EnableMediaReplace\Replacer->replaceWith()`. Since the hook `'wp_handle_replace'` is always used before creating a new instance of `Replacer`, it will work. This hook is also used by ShortPixel.

**Effort** ✅ 
Effort `[S]`

PR #494, opened 2020-05-06
Fixes #491: Auto-optimization on image upload doesn't optimize thumbnails

### LINKED ISSUE #491: Auto-optimization on image upload doesn't optimize thumbnails
Probably due to a new change in the upload process, only the full size (+ webp) is optimized automatically. The media then displays a "Optimize X missing thumbnails" link.
I was able to follow the lead to some point:
```
Imagify_Auto_Optimization->do_auto_optimization()
=> Imagify\Optimization\Process\AbstractProcess->optimize()
=> Imagify\Media\WP->get_media_files()
=> wp_get_attachment_metadata()
```

This final `wp_get_attachment_metadata()` returns an empty string instead of the full meta data (as it should, since it is supposed to be ran after the upload is complete).

Related [ticket](https://secure.helpscout.net/conversation/1129605795/156125?folderId=1724287).

Screenfeed:
Manual tests so far:
- Auto-optimize on upload in WP >= 5.3: ✅ 
- Auto-optimize on upload in WP < 5.3: ✅ 
- Compatibility with "Regenerate Thumbnails" still working (WP >= 5.3): ✅ 
- Compatibility with "Enable Media Replace" still working (WP >= 5.3): ❌ (but was broken before this PR, see #493).

Fixes #491.

### DIFF
--- a/inc/classes/class-imagify-auto-optimization.php
+++ b/inc/classes/class-imagify-auto-optimization.php
@@ -164,7 +164,7 @@ class Imagify_Auto_Optimization extends Imagify_Auto_Optimization_Deprecated {
 			return $metadata;
 		}
 
-		if ( $this->is_wp_53 && 'create' === $context ) {
+		if ( 'create' === $context ) {
 			$this->set_step( $attachment_id, 'upload' );
 		}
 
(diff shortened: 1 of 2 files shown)


Screenfeed:
As Tonya suggested, no automated tests yet: the plugin is not totally testable yet and the test suite may lack a few things, like filesystem. So, tests will be added later and we will focus on QA tests now so we can release quickly.

Screenfeed:
@wp-media/dev-qa Ready for QA.
[imagify-1.9.10-alpha1.zip](https://github.com/wp-media/imagify-plugin/files/4593303/imagify-1.9.10-alpha1.zip)

markonikolic985:
Tested this on WP v5.4.1 and WP v5.2.

All good, got the expected result in both.

PR #506, opened 2020-08-17
Remove whitelist filter, check for deprecated use

### LINKED ISSUE #505: Error: options page not found in Multisite (WP 5.5)
In WP 5.5 installation Multisite using subdomains:
Clicking "Save Changes" button from the Imagify > Settings page results in
`Error: options page not found` notice.

iCaspar:
Fixes #505 

WP 5.5 deprecates the `whitelist_options` filter. We have been using this filter in our options setting for multisite installs.

We are adding a check for the deprecated filter to maintain back-compatibility, then filtering also on the new WP filter: `allowed_options`.

Also, adding unit test coverage for `Imagify_Settings::update_site_option_on_network()`


Tests/Integration/inc/classes/ImagifySettings/updateSiteOptionOnNetwork.php
```diff
@@ -0,0 +1,112 @@
...
+	/**
+	 * @dataProvider provideData
+	 */
+	public function testShouldUpdateNetworkSettings( $config, $expected ) {
+		$_POST['option_page'] = $config['option_page'];
+
+		if ( $config['user_can'] ) {
+			$this->user_id = $this->factory->user->create( [ 'role' => 'administrator', ] );
+			$admin         = get_role( 'administrator' );
+			$admin->add_cap( 'manage_network_options' );
```
Tabrisrp, line 35: To prevent future conflicts between tests, you will want to reset the capability, & unset the `option_page` key on tearDown.

This can be applied to anything you set in the test method that is for a specific state, and which might not be accurate for another test.

Tests/Unit/inc/classes/ImagifySettings/updateSiteOptionOnNetwork.php
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		Functions\when( 'imagify_is_active_for_network' )->justReturn( true );
+	}
+
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldUpdateNetworkSettings( $config, $expected ) {
+		$_POST['option_page'] = $config['option_page'];
```
Tabrisrp, line 34: Here too you'll want to reset between tests on tearDown

Tests/Integration/inc/classes/ImagifySettings/updateSiteOptionOnNetwork.php
iCaspar: Done

Tests/Unit/inc/classes/ImagifySettings/updateSiteOptionOnNetwork.php
iCaspar: Done

iCaspar:
@wp-media/qa Ready for QA.

This is a multisite issue. For the regression with WP 5.5 see original issue, see #505 .
(It says "using subdomain" because that was the multisite environment in the presenting support ticket, and I replicated it that way -- but since the plugin uses the same method for any multisite, I believe it will have regressed on subdirectory versions as well.)

This branch also contains merged content from #508 So we'll want to QA for any regressions on single sites as well.

Davidacu:
- singlesite| WP 5.4.2 👍 
- singlesite| WP 5.5 👍 
- Multisite | subdomain | WP 5.5 👍 
- Multisite | subdirectory | WP 5.5 👍 
- Multisite | subdirectory | WP 5.4.2 🐞  
I was able to reproduce the issue in a local environment.
The error "ERROR: options page not found."  is displayed.

**Steps:**
1. Network activate the plugin
2. Go to Imagify settings and click the "Save changes" button.
image

if you activate the plugin from the sites then the issue is not reproducible.

iCaspar:
@wp-media/qa Ready for a recheck. Pls & Thank you!

PR #508, opened 2020-08-20
Update composer dependencies.

### LINKED ISSUE #507: Protect composer dependencies from collisions
Following Rocket issue 3003 (https://github.com/wp-media/wp-rocket/issues/3003), we are updating the way we manage composer dependencies in all our plugins to bullet-proof against future conflicts.

- [ ] Add and configure mozart to run during dev cycles only.
- [ ] Move production dependencies into `classes/Dependencies` & use `WPMIP` prefix.
- [ ] Update other dependencies to more current versions.
- [ ] Remove `composer.lock` from repo & ignore (generated in dev process)

iCaspar:
Closes #507 

Updates all dependencies to latest versions.
Adds mozart to re-namespace production dependencies.
Adds wp-media/phpunit package
Adds roave/security-advisories
Adds phpstan & phpstan-wordpress


.gitignore
```diff
@@ -2,6 +2,8 @@
 /node_modules
 /Tests/env/local
 /vendor
+/classes/Dependencies
```
Tabrisrp, line 5: The namespaced dependencies should be versioned, so that they are available when we publish the release via the Github action

composer.json
```diff
@@ -68,12 +73,32 @@
 	"autoload-dev": {
 		"psr-4": { "Imagify\\Tests\\": "Tests/"	}
 	},
+	"extra": {
+		"mozart": {
+			"dep_namespace": "ImagifyPlugin\\Dependencies\\",
```
Tabrisrp, line 78:
```suggestion
			"dep_namespace": "Imagify\\Dependencies\\",
```
Let's use the same namespace as our internals for consistency

composer.json
```diff
@@ -68,12 +73,32 @@
...
+			"dep_directory": "/src/Dependencies/",
```
Tabrisrp, line 79:
```suggestion
			"dep_directory": "/classes/Dependencies/",
```
Our namespaced files are in the `classes` directory, so we can use that for autoloading the dependencies too

composer.json
```diff
@@ -68,12 +73,32 @@
...
+			"classmap_directory": "/classes/Dependencies/",
```
Tabrisrp, line 80:
```suggestion
			"classmap_directory": "/inc/classes/Dependencies/",
```
Not-namespaced classses are in `inc/classes`, we can add the not-namespaced dependencies there too

composer.json
```diff
@@ -68,12 +73,32 @@
...
+			"classmap_prefix": "WPMIP_",
+			"packages": [
+				"deliciousbrains/wp-background-processing"
+			],
+			"delete_vendor_directories": true
```
Tabrisrp, line 85: The default  value is already `true` for this parameters, so we don't need to add it

.travis.yml
```diff
@@ -43,4 +64,12 @@ before_script:
...
-- composer run-tests
+    - |
+        if [[ "$WP_TRAVISCI" == "phpcs" ]]; then
+          composer phpcs
+        elif [[ "$WP_TRAVISCI" == "phpstan" ]]; then
+          composer require --dev szepeviktor/phpstan-wordpress
+          vendor/bin/phpstan analyze
+        else
+          composer run-tests
+        fi
```
Tabrisrp, line 75: You'll run into failures with the addition of Mozart, you can have a look at what I did there to keep Travis running as expected: https://github.com/wp-media/wp-rocket/pull/3008/files#diff-354f30a63fb0907d4ad57269548329e3

inc/classes/Dependencies/deliciousbrains/wp-background-processing/classes/wp-background-process.php
```diff
@@ -0,0 +1,505 @@
...
+ * @package WP-Background-Processing
+ */
+
+/**
+ * Abstract WP_Background_Process class.
+ *
+ * @abstract
+ * @extends WP_Async_Request
+ */
+abstract class Imagify_WP_Background_Process extends WP_Async_Request {
```
Tabrisrp:
```suggestion
abstract class Imagify_WP_Background_Process extends Imagify_WP_Async_Request {
```

PR #513, opened 2020-09-02
Adds an empty index file when creating directories

iCaspar:
Adding index files to directories we create in the filesystem to short-circuit public access to contents on servers that allow indexes.

## TODO

- [x] Create indexes in new directories as they are created.
- [x] Run a directory scan of previously created Imagify directories on update and add index files where missing.
  - [X] Recursive Scan and add to `imagify-backup` directory in WP root.
  - [X] Recursive Scan and add to `wp-content/uploads/backup`.
  - [X] Scan and add to `conf` directory in WP root.
- [x] Validate with Next Galley plugin to ensure it properly creates the `index.php` file for each newly created directory
- [x] Add unit tests
- [x] Add integration tests
(See note below re: unit and integration tests.)


iCaspar:
On the "Validate with NextGEN" item in the OP, we do not create separate backup directories for images uploaded into NextGEN galleries. Rather, backups are stored in directories already created by NextGEN alongside the optimized ones. Since we don't create those, we're going to leave them for NextGen to manage.

iCaspar:
@wp-media/qa @Davidacu This is ready for QA, Round 1.

Davidacu (commented):
- Please find all scenarios tested under the test execution results file attached bellow
- The failure is unrelated to the changes, I already asked the team for more information.
- There is an index.php file being created under `/wp-content/uploads/` directory, it was fixed before but now it seems to be added back, could you please check @iCaspar 
[testrail-report-26.pdf](https://github.com/wp-media/imagify-plugin/files/5171856/testrail-report-26.pdf)

Davidacu (approved):
- All issues have been fixed.
- All cases have passed, please find the test result attached.
[testrail-report-27.pdf](https://github.com/wp-media/imagify-plugin/files/5176183/testrail-report-27.pdf)

iCaspar:
Unit and integration tests have been added here, but will require additional coverage in future releases. Because of the old architecture, adding further unit/integration coverage at this point will require major refactoring, too extensive to include in a minor release.
@hellofromtonya , This has passed QA is ready for code review.

inc/classes/class-imagify-custom-folders.php
```diff
@@ -89,6 +89,49 @@ public static function get_file_backup_path( $file_path ) {
...
+					continue;
+				}
+
+				$path = $fileinfo->getRealPath();
+
+				if ( ! $filesystem->is_file( trailingslashit( $path ) . 'index.html' )
+					&& ! $filesystem->is_file( trailingslashit( $path ) . 'index.php' )
+				) {
+					$filesystem->touch( trailingslashit( $path ) . 'index.php' );
+				}
```
hellofromtonya, line 129: Apply `trailingslashit( $path )` before this block to do it once instead of 3x.

imagify.php
```diff
@@ -3,7 +3,7 @@
  * Plugin Name: Imagify
  * Plugin URI: https://wordpress.org/plugins/imagify/
  * Description: Dramaticaly reduce image file sizes without losing quality, make your website load faster, boost your SEO and save money on your bandwidth using Imagify, the new most advanced image optimization tool.
- * Version: 1.9.10
+ * Version: 1.9.11
```
hellofromtonya, line 6: Please add the following header fields:

- Requires at least: The lowest WordPress version that the plugin will work on.
- Requires PHP: The minimum required PHP version.

[See WordPress Plugin manual for more info](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields).

hellofromtonya (changes requested):

hellofromtonya (approved):

PR #523, opened 2020-10-30
Enhancement/pricing use active plans

iCaspar:
Update plugin to handle new pricing structure and pricing api json responses:
see issue https://github.com/wp-media/imagify/issues/1128 for more background and criteria.
Closes https://github.com/wp-media/imagify/issues/1128

- [x] Handle when no onetime plans are available
- [x] Handle new plan active status to sort which plans to display.
- [x] Work with current (pre 11.20) pricing api json response.
- [x] Work with new (11.20) pricing api json response.


engahmeds3ed (approved):

piotrbak:
I tested everything on the **old** API, the JS code works:
- All plans are displayed correctly
- When plans are not active, they're not displayed (happens just now)
- The plan names are displayed correctly
- We're able to buy different plans
- Size calculations work just fine

There are a couple of things I'm having problems with, they don't look like regression though:
-  After adding the card for the first time, I'm not able to make another purchase (upgrade the license) anymore. Maybe it's related to the `dapp` environment. The JS error:
https://snippi.com/s/a4qy13d
- I don't have the possibility to remove the existing cards from the system.
- The popin doesn't work on Chrome
- After choosing the plan and clicking on the `checkout`, it's not possible to close the window with the `x`. Maybe it's related to the `dapp` environment. Not able to test on the live, since plans are not active.
- The dashboard license text near the `Account Type` doesn't refresh until the transients are cleared. Do we clear the transients after the purchase is made?

engahmeds3ed (approved):

PR #530, opened 2020-11-13
Migrate from Travis to GA

AlbertCintas:


remyperona:
@AlbertCintas I went back on this to finalize, there is one remaining issue I'm not able to solve:

On PHP 5.6 only, the workaround command for MySQL doesn't seem to take effect. The other PHP versions are OK.

If you can have a look? Else we will just ditch the test for 5.6.

remyperona (approved):

PR #540, opened 2020-12-16
Fix PHP8-cURL Compatibility

### LINKED ISSUE #539: PHP8 and cURL compatibility
PHP 8.0 changes the type returned from `curl_init()` from `resource` to a `CurlHandle` object.
(see https://www.php.net/manual/en/function.curl-init.php)

We're checking for a resource [here](https://github.com/wp-media/imagify-plugin/blob/9d88d5d4953044b8e53ba71e065e138b16bc8d44/inc/classes/class-imagify.php#L473), which results in errors when sending cURL requests with images.

We need to also check for the new object before failing with an error notice.

iCaspar:
Fixes #539 .

Add a check for php8 CurlHandle object

### DIFF
--- a/inc/classes/class-imagify.php
+++ b/inc/classes/class-imagify.php
@@ -470,7 +470,7 @@ class Imagify {
 			$url = self::API_ENDPOINT . $url;
 			$ch  = @curl_init();
 
-			if ( ! is_resource( $ch ) && 'object' !== gettype( $ch ) && 'CurlHandle' !== get_class( $ch ) ) {
+			if ( ! is_resource( $ch ) && ( 'object' !== gettype( $ch ) && 'CurlHandle' !== get_class( $ch ) ) ) {
 				throw new Exception( 'Could not initialize a new cURL handle' );
 			}
 
--- a/inc/classes/class-imagify.php
+++ b/inc/classes/class-imagify.php
@@ -468,7 +468,7 @@ class Imagify {
 
 		try {
 			$url = self::API_ENDPOINT . $url;
-			$ch  = @curl_init();
+			$ch  = curl_init();
 
 			if ( false === $ch ) {
 				throw new Exception( 'Could not initialize a new cURL handle' );
--- a/inc/classes/class-imagify.php
+++ b/inc/classes/class-imagify.php
@@ -470,7 +470,7 @@ class Imagify {
 			$url = self::API_ENDPOINT . $url;
 			$ch  = @curl_init();
 
-			if ( ! is_resource( $ch ) || 'CurlHandle' !== gettype( $ch ) ) {
+			if ( ! is_resource( $ch ) && 'object' !== gettype( $ch ) && 'CurlHandle' !== get_class( $ch ) ) {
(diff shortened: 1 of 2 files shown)


inc/classes/class-imagify.php
```diff
@@ -470,7 +470,7 @@ private function curl_http_call( $url, $args = [] ) {
 			$url = self::API_ENDPOINT . $url;
 			$ch  = @curl_init();
 
-			if ( ! is_resource( $ch ) ) {
+			if ( ! is_resource( $ch ) && ( ! is_object( $ch ) && 'CurlHandle' !== get_class( $ch ) ) ) {
```
hellofromtonya, line 473: Could use [`instanceof`](https://www.php.net/manual/en/language.operators.type.php) here to compress the code as it'll check if the `$ch` is an object and an instance of the class.

iCaspar: @engahmeds3ed suggested that simply checking for `false` works for any php version here. Even simpler! Props!

inc/classes/class-imagify.php
```diff
@@ -470,7 +470,7 @@ private function curl_http_call( $url, $args = [] ) {
 			$url = self::API_ENDPOINT . $url;
 			$ch  = @curl_init();
```
hellofromtonya, line 471: Note: In PHP 8, `@` will no longer silence fatal errors. [See the notes here](https://php.watch/versions/8.0/fatal-error-suppression).

iCaspar: As nearly as I  can tell, the error suppression here would have been to avoid a fatal error when `curl_init()` was not available. From v1.7.1, there is a separate check, `static supports_curl()` added to `Imagify_Requirements` class to throw an exception for that possibility, so the `@` suppression is redundant here in any version. Any other errors from `curl_init()` itself will cause it to return `false` which we're now going to check for (again, works in any version).

iCaspar:
@wp-media/qa - Acceptance criteria:
1. Users of Imagify on PHP 8 installation should not get a `Could not initialize a new cURL handle` error when attempting to optimize images.
2. Users of Imagify on PHP < 8 should not experience any changes.
3. Regression test: We have removed what we believe to be a redundant error supression to handle when cURL is not available at all. In this case, users should see a `cURL isn\'t installed on the server.` error message in the Media Library (or other relevant) UI rather than a PHP fatal error. This should be regression-checked.

hellofromtonya (approved):

UltraFlux:
confirmed working, updated and tested
Server Info : nginx/1.19.6
PHP Version : 8.0.0
Jnews 7.1.x

piotrbak (approved):
Worked as expected. Tested scenarios:
- PHP8 image optimization 👍
- PHP8 bulk optimization 👍

- PHP7.4 image optimization 👍
- PHP7.4 bulk optimization 👍
- PHP7.4 cURL disabled 👍 ( Message displayed when doing bulk optimization, when doing a single file optimization, no message is present, no error in the logs too. Looped screen only below )
image

PR #558, opened 2021-04-08
Update api interactions with imagify.io

iCaspar:
A few updates to reflect recent changes on the imagify.io api:

1. Send the current coupon code to the app's api in order to correctly calculate pricing on the app (inside the pricing modal's iframe).
2. Update the api endpoints for development and testing environments.
3. ~Handle a regression where coupon discounts were not appearing after selection of a new discounted plan from the "view all plans" tab of the pricing modal.~


iCaspar:
After conversation with @piotrbak we're going to narrow this PR to just 1 and 2, above. We'll take care of 3 separately, as there are several issues with the modal window's behavior that need to be addressed in consultation with product and QA.

iCaspar:
Closing this and replacing with #585. Plugin is sending coupon data correctly already in the payment iframe's src url, so sending in another data object is redundant here.

Issue #557, opened 2021-04-08
Coupon code is automatically applied to all plans in the UI, even to those that are not discounted


piotrbak:
**Is your feature request related to a problem? Please describe.**
When the promo is applied only to the specific plan, the coupon code and the label `20% off with spring2021` is visible all the time:
![image](https://user-images.githubusercontent.com/26767037/114056288-6b1df900-9889-11eb-8202-b68c08588748.png)
Because of that, we're not able to apply another coupon to different, not discounted plans.

**Describe the solution you'd like**
I think that the coupon + label should be added only when the chosen plan is discounted in the promo. We should be also able to edit this field and apply different coupon

piotrbak:
@Tabrisrp The promo is off, but this change is essential for the new promo. It prevents from buying another plan:
![image](https://user-images.githubusercontent.com/26767037/114692825-e3b10980-9d10-11eb-938f-47453a964ddb.png)

It's happening only on the newest branch that we are working on currently. Before the coupon wasn't applied so checkout worked for all plans.

iCaspar:
The pop-over shown there is being generated from inside the app, and is to do with the handling of the coupon mismatch there. If there is a way to resolve that more gracefully there, we should.
However, on the plugin side, we should implement a UI of our own that will redirect customers from applying a coupon on an order where it is not valid **before** allowing them to proceed to the checkout page.

PR #577, opened 2021-07-06
fix phpcs 7.4 errors

mostafa-hisham:
i fixed all phpcs errors for php 7.4
also I removed _wp_json_prepare_data function
because it is deprecated from WordPress version 5.3
it was handling a case in PHP 5.2 and 5.3, Imagify plugin requires PHP 5.4 to work so there is no need for it
and it is removed from the WPcore  2years ago
https://core.trac.wordpress.org/ticket/47699

Also, the WordPress min requirement is PHP version is 5.6.2 and I think we should make  Imagify plugin require ph 5.6 or 5.8 to work to keep up with WordPress.
https://wordpress.org/about/requirements/

`Note: If you are in a legacy environment where you only have older PHP or MySQL versions, WordPress also works with PHP 5.6.20+ and MySQL 5.0+, but these versions have reached official End Of Life and as such may expose your site to security vulnerabilities.`

### DIFF
--- a/classes/Webp/Picture/Display.php
+++ b/classes/Webp/Picture/Display.php
@@ -193,13 +193,13 @@ class Display {
 		 * @param array $data       Data built from the originale <img> tag. See $this->process_image().
 		 */
 		$attributes = apply_filters( 'imagify_picture_attributes', $attributes, $image );
-		
+
 		/**
 		 * Remove Gutenberg specific attributes from picture tag, leave them on img tag.
 		 * Optional: $attributes['class'] = 'imagify-webp-cover-wrapper'; for website admin styling ease.
 		 */
 		if ( ! empty( $image['attributes']['class'] ) && strpos( $image['attributes']['class'], 'wp-block-cover__image-background' ) !== false ) {
-        	unset( $attributes['style'] );
+			unset( $attributes['style'] );
 			unset( $attributes['class'] );
 			unset( $attributes['data-object-fit'] );
 			unset( $attributes['data-object-position'] );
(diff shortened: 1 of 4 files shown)


iCaspar (approved):

iCaspar:
On minimum version requirements. I think we can safely remove the legacy function that has been removed from core itself. That said, we are currently advertising on the wpOrg repo that [minimums are WP 4.0.0 / PHP 5.4](https://github.com/wp-media/imagify-plugin/blob/develop/imagify.php#L7-L8).

@Tabrisrp @GeekPress @Agathemed We should probably update the published minimum requirements with this release, given these changes -- and that might also imply that the next release should indicate possible breaking changes, so maybe revise release version to 1.10.0. Thoughts?

remyperona:
I'm ok with that, I think we should bump the versions to WP 5.3 and PHP 7.0 to match with what we are using for WP Rocket

iCaspar:
@wp-media/qa Ready for QA. Mainly looking for regressions on this one, but this is no longer expected to work on versions below what is mentioned above.

piotrbak:
Did basic regression for this PR, it looks good:
[testrail-report-212.pdf](https://github.com/wp-media/imagify-plugin/files/6785550/testrail-report-212.pdf)

There's important notice for @mostafa-hisham as he's working on https://github.com/wp-media/imagify-plugin/issues/557, we talked about this today on standup:

**Without PR577:**
- The plans are crossed out when choosing them, they shouldn't be as the promo is only for Unlimited. (https://jmp.sh/JC1DY92)
- When the plan is chosen, then the price is crossed out, while it shouldn't be. (note the price there, it's CORRECT, https://jmp.sh/wF40rCe)
- It's not possible to change the promo coupon

**With PR577:**
- The plans are not crossed out when choosing them https://jmp.sh/X0ZWmDJ)
- When the plan is chosen, then the price is crossed out, while it shouldn't be. (note the price there, it's WRONG, https://jmp.sh/iJOpoHz)
- It's not possible to change the promo coupon

piotrbak (approved):

PR #578, opened 2021-07-06
Fix #564 - Improve filter for Imagify Notices

### LINKED ISSUE #564: Add a filter on Imagify notices

We currently have no way to remove notices displayed by Imagify. For an automatic installation with some of our partners, we need to remove the registration banner.

We are managing the display for all notices based on `all_admin_notices ` hook:
https://github.com/wp-media/imagify-plugin/blob/5971929e9842066e1351704ce14f25c5f31a7180/inc/classes/class-imagify-notices.php#L147

iCaspar:
Closes #564 

Per grooming notes, there is already a filter. It is suggested adding a cast to array before returning from `get_notice_ids()`.

The filter/method has been around since 1.6 without the cast, but the filters @param has indicated that an array is expected.

### DIFF
--- a/inc/classes/class-imagify-notices.php
+++ b/inc/classes/class-imagify-notices.php
@@ -872,7 +872,7 @@ class Imagify_Notices extends Imagify_Notices_Deprecated {
 		 *
 		 * @param array $notice_ids An array of notice "IDs".
 		 */
-		return apply_filters( 'imagify_notices', self::$notice_ids );
+		return (array) apply_filters( 'imagify_notices', self::$notice_ids );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


remyperona (approved):

iCaspar:
Ready for @wp-media/qa 
Should be able to change notices by hooking to `imagify_notices`.

piotrbak:
@iCaspar How can I remove the registration banner using this specific filter? I'd like to test scenario provided in the issue.

piotrbak (approved):
Working good! We should somewhere document what values can be used there :) 

[testrail-report-214.pdf](https://github.com/wp-media/imagify-plugin/files/6801621/testrail-report-214.pdf)

PR #583, opened 2021-07-07
Closes #561 Centralize App URL definition

### LINKED ISSUE #561: URL for the app API should be defined in one place.
The URL for reaching the API is defined in (at least) 3 different places.

`inc/functions/common.php`
`inc/classes/class-imagify.php`
`assets/js/pricing-modal.js`

This should be defined once and reused.

mostafa-hisham:
define app url and use it i PHP and js


inc/classes/class-imagify.php
```diff
@@ -12,7 +12,7 @@ class Imagify {
 	 *
 	 * @var string
 	 */
-	const API_ENDPOINT = 'https://app.imagify.io/api/';
+	const API_ENDPOINT = IMAGIFY_APP_DOMAIN.'/api/';
```
iCaspar:
```suggestion
	const API_ENDPOINT = IMAGIFY_APP_DOMAIN . '/api/';
```

inc/functions/common.php
```diff
@@ -223,8 +223,8 @@ function imagify_can_optimize_custom_folders() {
  * @return string The URL.
  */
 function imagify_get_external_url( $target, $query_args = array() ) {
-	$site_url = 'https://imagify.io/';
-	$app_url  = 'https://app.imagify.io/#/';
+	$site_url = IMAGIFY_APP_DOMAIN.'/';
```
iCaspar:
```suggestion
	$site_url = IMAGIFY_APP_DOMAIN . '/';
```

inc/functions/common.php
```diff
@@ -223,8 +223,8 @@ function imagify_can_optimize_custom_folders() {
...
+	$app_url  = IMAGIFY_APP_DOMAIN.'/#/';
```
iCaspar:
```suggestion
	$app_url  = IMAGIFY_APP_DOMAIN . '/#/';
```

iCaspar (changes requested):
Looks good, except these spacings will trigger phpcs even after the 7.4 update.

inc/functions/common.php
mostafa-hisham: fixed

inc/functions/common.php
mostafa-hisham: fixed

inc/classes/class-imagify.php
mostafa-hisham: this one generated an error in php cs
`Constant scalar expressions are not allowed when defining constants using the const keyword in PHP 5.5 or earlier.`
so I added new constant for API and used it in this class

https://github.com/wp-media/imagify-plugin/blob/5d82893ac318aae56da21bc3740469ec7dda4b17/imagify.php#L33

iCaspar (approved):

iCaspar:
Ready for @wp-media/qa 

This will resolve the issue about the pricing modal closing in test envs. The app URI to test for various environments is now in `imagify.php` -- set and forget! 🤞

piotrbak:
@iCaspar @mostafa-hisham What would you say about defining it like we do in WP Rocket?
https://github.com/wp-media/wp-rocket/blob/a94306216879962bdee8736fb93fd0f38b5de8da/wp-rocket.php#L60

It would allow us to define it in `wp-config.php` and make things even easier (switching between branches, etc.)

mostafa-hisham:
@piotrbak
Great idea 
Now we check first if it is not defined on wp-config.php
please check my changes

piotrbak:
@iCaspar @mostafa-hisham The JS change should make the payment modal to be closed automatically after the successful checkout, right?

If that's the case, then it's not working.

PR #585, opened 2021-07-07
Fix #584 Close Pricing Modal when using with test envs

### LINKED ISSUE #584: Payment Modal does not close when using plugin with test invironments.
- Made sure you’re on the latest version ✅ 1.9.14
- Used the search feature to ensure that the bug hasn’t been reported before ✅ 

**Describe the bug**
When using the plugin with our staging and test versions of the app, the pricing modal does not close after completing a purchase transaction.

**To Reproduce**
1. Point the app api URL to one of the test app URIs
2. Complete a test purchase in the plugin's pricing/checkout modal.
3. See the modal does not close after the transaction completes.

**Expected behavior**
The modal should close correctly in conjunction any app environment.

**Additional Context**
This issue is the only issue left from #558, which was trying to solve issues around the pricing modal that arose on-the-fly during the 2021 spring campaign.

- [X] Reproduce the problem
- [X] Identify the root cause -- the test and development URIs are not set correctly in the pricing modal js.
- [X] Scope a solution -- add the correct URIs into the `checkPluginMessage()` method in `pricing-modal.js`
- [X] Estimate the effort: [XS]

iCaspar:
Closes #584 

Per grooming, we're just updating the check against URLs to include current test app envs.


mostafa-hisham:
this [PR](https://github.com/wp-media/imagify-plugin/pull/583)  will fix this issue but in a different way
because we will define the URL in one place and send it to js 
we don't need to add multiple cases for each environment URL

iCaspar:
Per @mostafa-hisham comment above, closing this.

PR #587, opened 2021-07-07
Fixes #495 Webp compatibility with WooCommerce variable product images

### LINKED ISSUE #495: WebP (picture tags) does not work with WooCommerce variable products
It looks like using WebP display (picture tags) won't work fine with WooCommerce variable products.

Tickets are tagged "img_variations" on HS - https://secure.helpscout.net/search/?query=tag%3Aimg_variations

iCaspar:
Closes #495 

This is similar to what was groomed, but the method for invoking the filters is in a new compatibility class, and the filters themselves are more careful to guard against inadvertently removing other classes or adding/overwriting classes to other images on the page than the proof-of-concept code given.

### DIFF
--- a/inc/3rd-party/WooCommerce/WooCommerce.php
+++ b/inc/3rd-party/WooCommerce/WooCommerce.php
@@ -1,0 +1,58 @@
+<?php
+namespace Imagify\ThirdParty\WooCommerce;
+
+class WooCommerce {
+	public function init() {
+		add_action( 'woocommerce_single_product_summary', [ $this, 'variable_products_webp_compat' ] );
+	}
+
+	public function variable_products_webp_compat() {
+		global $product;
+
+		if ( ! isset( $product ) || ! $product->is_type( 'variable' ) ) {
+			return;
+		}
+
+		add_filter( 'imagify_picture_attributes', [ $this, 'remove_wp_post_image_class' ], 10, 2 );
+		add_filter(
+			'imagify_picture_source_attributes',
+			[ $this, 'maybe_add_wp_post_image_class_on_picture_internal_tags' ],
+			10,
+			2
+		);
+		add_filter(
+			'imagify_picture_img_attributes',
+			[ $this, 'maybe_add_wp_post_image_class_on_picture_internal_tags' ],
+			10,
+			2
+		);
+	}
+
+	function remove_wp_post_image_class( $attributes, $image ) {
+		if ( isset( $attributes['class'] ) ) {
+			$attributes['class'] = str_replace( 'wp-post-image', '', $attributes['class'] );
+		}
+
+		if ( empty( $attributes['class'] ) ) {
+			unset( $attributes['class'] );
+		}
+
+		return $attributes;
+	}
+
+	function maybe_add_wp_post_image_class_on_picture_internal_tags( $attributes, $image ) {
+		if (
+			! empty( $image['attributes']['class'] )
+			&& strpos( $image['attributes']['class'], 'wp-post-image' ) !== false
+		) {
+			$attributes['class'] = isset( $attributes['class'] )
+				? $attributes['class'] . ' wp-post-image'
+				: 'wp-post-image';
+		}
+
+		return $attributes;
+	}
+}
+
+$imagify_woocommerce_compat = new WooCommerce();
+$imagify_woocommerce_compat->init();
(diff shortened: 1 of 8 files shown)


mostafa-hisham (changes requested):
php cs generated some errors for this class

- Filenames should be all lowercase with hyphens as word separators. Expected woocommerce.php, but found WooCommerce.php.
- Class file names should be based on the class name with "class-"
- Missing doc comment for functions and the class
- Visibility (Access Modifiers) must be declared on methods

inc/3rd-party/WooCommerce/WooCommerce.php
```diff
@@ -0,0 +1,58 @@
...
+				? $attributes['class'] . ' wp-post-image'
+				: 'wp-post-image';
+		}
+
+		return $attributes;
+	}
+}
+
+$imagify_woocommerce_compat = new WooCommerce();
+$imagify_woocommerce_compat->init();
```
remyperona:
```suggestion
( new WooCommerce() )->init();
```

inc/3rd-party/WooCommerce/WooCommerce.php
```diff
@@ -0,0 +1,58 @@
...
+		}
+
+		if ( empty( $attributes['class'] ) ) {
+			unset( $attributes['class'] );
+		}
+
+		return $attributes;
+	}
+
+	function maybe_add_wp_post_image_class_on_picture_internal_tags( $attributes, $image ) {
```
remyperona:
```suggestion
	public function maybe_add_wp_post_image_class_on_picture_internal_tags( $attributes, $image ) {
```

inc/3rd-party/WooCommerce/WooCommerce.php
```diff
@@ -0,0 +1,58 @@
...
+		);
+		add_filter(
+			'imagify_picture_img_attributes',
+			[ $this, 'maybe_add_wp_post_image_class_on_picture_internal_tags' ],
+			10,
+			2
+		);
+	}
+
+	function remove_wp_post_image_class( $attributes, $image ) {
```
remyperona:
```suggestion
	public function remove_wp_post_image_class( $attributes, $image ) {
```

remyperona (changes requested):
Missing docblocks everywhere, but validated the logic

iCaspar:
Indeed, phpcs will not be happy with this yet.

Most of the rest of the plugin is using static/singleton type classes (a common WP plugin practice).
However, instead of using that pattern, I've put this in as a class and (another common WP plugin practice) created an instance and initialized it from within the included class file.

The other alternative I considered was folding these methods into the Webp/Picture/Display class -- but that does not properly separate concerns.

Per @Tabrisrp this will do until we have a chance to re-architect more thoroughly. I'll fix to conform to the naming conventions, etc.

iCaspar:
@wp-media/php @wp-media/qa Ready for recheck and QA.

remyperona (approved):

piotrbak:
@iCaspar Could you tell me what is exactly going to happen in this PR? Are we going to add `wp-post-image` class to the images?

iCaspar:
So, on this one, WooCommerce is targeting images with `wp-post-image` class in the script that enables swapping out product images when the visitor selects different product variations from the drop-down on variable products selection. BUT the library does not support picture tags. To help them target the images for swap, we need to (1) remove `wp-post-image` class from the `<picture>` tag, and (2) add the `wp-post-image` class to all the `<source>` and `` tags inside the `<picture>` tag.

In HTML inspector we should see:
Before:
```
<picture class="wp-post-image"...>
   <source src="path/to/image.webp"....>
   <img src="path/to/image.jpg"...>
</picture>
```
After:
```
<picture ...> // no class or class without wp-post-image included
   <source class="wp-post-image" src="path/to/image.webp"....>
   <img class="wp-post-image" src="path/to/image.jpg"...>
</picture>
```

Also, this should be done _only_ to images that are inside the selection/description part of the page, leaving other images on the page untouched. And it should happen _only_ on variable products.

And, the image should change as expected based on selection of variation :)

piotrbak (approved):
Looks like the problem is fixed, good to go!

[testrail-report-215.pdf](https://github.com/wp-media/imagify-plugin/files/6802613/testrail-report-215.pdf)

PR #590, opened 2021-07-08
Fix #537 Stop replacing images already nested inside picture tags

### LINKED ISSUE #537: Imagify replaces img tags inside existing picture tags resulting in nested picture tags
**Describe the bug**
When webp images display option with picture tags is used from the settings and theme already has picture tags present, then Imagify replaces img tags inside existing picture tags resulting in nested <picture> tags.

**Expected behavior**
Imagify should ignore already existing picture tags with webp images.

**Screenshots**
Screenshot_30 (1)

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

iCaspar:
Closes #537

Starting out adding a remove_picture_tags() method, per grooming notes.

~However, we cannot remove them without later putting them back at the end of the process, since we're calling `process_content()` inside of `ob_start()`. When we flush the buffer without putting the removed things back, they'll be missing from the page!~

~To remedy this, we'll use `preg_replace_callback()` in the remove part to make a placeholder comment in the html where we've removed a tag and store all the picture tags we've removed; then we can `preg_replace_callback()` again in a `reinsert_picture_tags()` method to swap the original picture tag back in at each placeholder.~

### DIFF
--- a/classes/Webp/Picture/Display.php
+++ b/classes/Webp/Picture/Display.php
@@ -32,6 +32,14 @@ class Display {
 	protected $filesystem;
 
 	/**
+	 * Pre-existing picture tags extracted html.
+	 *
+	 * @var array
+	 * @since 1.10.0
+	 */
+	private $picture_tags;
+
+	/**
 	 * Constructor.
 	 *
 	 * @since  1.9
@@ -148,7 +156,7 @@ class Display {
 			$content = str_replace( $image['tag'], $tag, $content );
 		}
 
-		return $content;
+		return $this->reinsert_picture_tags( $content );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/Webp/Picture/Display.php
```diff
@@ -31,6 +31,14 @@ class Display {
 	 */
 	protected $filesystem;
 
+	/**
+	 * Pre-existing picture tags extracted html.
+	 *
+	 * @var array
+	 * @since 1.10.0
+	 */
+	private $picture_tags;
```
remyperona:
```suggestion
	private $picture_tags = [];
```

classes/Webp/Picture/Display.php
```diff
@@ -147,7 +156,51 @@ public function process_content( $content ) {
...
+	 * Used in conjunction with remove_picture_tags() to re-construct HTML after having replaced non-nested images.
+	 *
+	 * @since 1.10.0
+	 *
+	 * @param string $html Content of the page without pre-existing <picture> tags.
+	 *
+	 * @return string HTML content with pre-existing <picture> tags re-inserted.
+	 */
+	private function reinsert_picture_tags( $html ) {
+		return preg_replace_callback(
```
remyperona: I'd suggest adding a bailout condition here to return early the `$html` if `$this->picture_tags` is empty

remyperona (changes requested):

mostafa-hisham:
@iCaspar 
Sorry for not clarifying this in the grooming

what I wanted to do is remove the `<picture>` tags first then use this HTML(without picture tag) to find the rest of the `` tags
then use the original HTML content in the rest of the process

will be like this in code
```
	public function process_content( $content ) {
		$html_no_picture_tags= $this->remove_picture_tags($content);
		$images = $this->get_images( $html_no_picture_tags );

		if ( ! $images ) {
			return $content;
		}

		foreach ( $images as $image ) {
			$tag     = $this->build_picture_tag( $image );
			$content = str_replace( $image['tag'], $tag, $content );
		}

		return $content;
	}
```
it is the same solution we used  in RUCSS to avoid styles in `<noscript>` tags
https://github.com/wp-media/wp-rocket/blob/2d094e01278e3e911d50cdded013f22f53ea4e70/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L427-L479

I think the solution we used in RUCSS is more simple

classes/Webp/Picture/Display.php
iCaspar: Great Idea!

remyperona (approved):

iCaspar:
Ready for @wp-media/qa

piotrbak (approved):
We're good here too!

[testrail-report-213.pdf](https://github.com/wp-media/imagify-plugin/files/6792973/testrail-report-213.pdf)

PR #598, opened 2021-07-13
Fix #573 - PHP Notice when user has no image count

### LINKED ISSUE #573: PHP Notice related to $image_count
**Describe the bug**
I'm getting this PHP notice on my local installation with Imagify active:
```
PHP Notice:  Undefined property: stdClass::$image_count in /imagify-plugin/inc/classes/class-imagify-cron-rating.php on line 106
[18-Jun-2021 16:10:30 UTC] PHP Stack trace:
[18-Jun-2021 16:10:30 UTC] PHP   1. {main}() /wp-cron.php:0
[18-Jun-2021 16:10:30 UTC] PHP   2. do_action_ref_array() /wp-cron.php:138
[18-Jun-2021 16:10:30 UTC] PHP   3. WP_Hook->do_action() /wp-includes/plugin.php:551
[18-Jun-2021 16:10:30 UTC] PHP   4. WP_Hook->apply_filters() /wp-includes/class-wp-hook.php:316
[18-Jun-2021 16:10:30 UTC] PHP   5. Imagify_Cron_Rating->do_event() /wp-includes/class-wp-hook.php:292
```

Not a big issue, but we should avoid it. I'm guessing it's happening because I have the plugin activated but didn't optimize any image, so there is no value for this property. We can guard that with an `isset()` or something.

**Expected behavior**
No PHP notice

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

iCaspar:
Closes #573 
Per grooming, just adds the isset() check.

### DIFF
--- a/inc/classes/class-imagify-cron-rating.php
+++ b/inc/classes/class-imagify-cron-rating.php
@@ -103,7 +103,7 @@ class Imagify_Cron_Rating extends Imagify_Abstract_Cron {
 
 		$user = get_imagify_user();
 
-		if ( ! is_wp_error( $user ) && (int) $user->image_count > 100 ) {
+		if ( ! is_wp_error( $user ) && isset( $user->image_count ) && (int) $user->image_count > 100 ) {
 			set_site_transient( 'imagify_user_images_count', $user->image_count );
 		}
 	}


remyperona (approved):

piotrbak (approved):
Didn't reproduce it, but setting the `imagify_user_images_count` transient works just fine with no regressions.

PR #601, opened 2021-07-13
Fix #565 - Array Flip fails on non-string item

### LINKED ISSUE #565: PHP Warning:  array_flip(): Can only flip STRING and INTEGER values! in /public_html/wp-content/plugins/imagify/inc/classes/class-imagify-admin-ajax-post.php on line 471
**Describe the bug**
An Imagify user's error log showed the following warning related to the Bulk Optimizer stats:

`PHP Warning:  array_flip(): Can only flip STRING and INTEGER values! in /public_html/wp-content/plugins/imagify/inc/classes/class-imagify-admin-ajax-post.php on line 471`

**To Reproduce**
Not entirely sure, as I'm not sure what is specifically causing this, but for this user, the Warning is logged when the Bulk Optimizer page is viewed.

**Expected behavior**
This warning message should not be caused when viewing the Bulk Optimizer page.

We could use the `double` `array_flip()` trick or `array_unique()` to resolve this as was done in this [WP Rocket Issue.](https://github.com/wp-media/wp-rocket/issues/3238)

**Additional Context**
Related ticket: https://secure.helpscout.net/conversation/1518484990/265351/

iCaspar:
Closes #565 

Per grooming / discussion on 565: adds a check for is_string before flipping.

### DIFF
--- a/inc/classes/class-imagify-admin-ajax-post.php
+++ b/inc/classes/class-imagify-admin-ajax-post.php
@@ -454,7 +454,7 @@ class Imagify_Admin_Ajax_Post extends Imagify_Admin_Ajax_Post_Deprecated {
 		imagify_check_nonce( 'imagify-bulk-optimize' );
 
 		$folder_types = filter_input( INPUT_GET, 'types', FILTER_SANITIZE_STRING, FILTER_REQUIRE_ARRAY );
-		$folder_types = is_array( $folder_types ) ? array_filter( $folder_types ) : [];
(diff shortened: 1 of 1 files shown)


remyperona (approved):

iCaspar:
Ready for @wp-media/qa

piotrbak (approved):
@iCaspar I also wasn't able to reproduce the error... However, I tested the Bulk Optimziation page and there were no regressions.

PR #626, opened 2021-11-01
Fix/603 small css UI glitch

### LINKED ISSUE #603: Small CSS UI glitch

**Describe the bug**
The ` You have used all credits` button is too big when the popup is opened, there should be better padding used on the texts too

**To Reproduce**
1. User 100% of your quota
2.Go to the Bulk Optimization
3. Try to optimize

**Expected behavior**
Better responsivness

**Screenshots**
image

**Additional context**

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

joejoe04:
closes #603 

Fixes issues where the overquota-alert title's line-height was too small and the button was cut off at narrow widths:

Also fixes a small typo in the button's text:


iCaspar (approved):

Mai-Saad (approved):
The main issue here is fixed (checked on different devices ios/android and different browser Chrome, Safari) can see other UI problems (will add a separate issue for that)
Note:
using chrome mobile view on a laptop, sometimes the red height > that on the device (using browserstack) as here 
Screenshot from 2021-11-03 12-05-29

PR #627, opened 2021-11-02
fix/594 ui problem with progress bar bulk optimization section

### LINKED ISSUE #594: UI problem with progress bar in the Bulk Optimization section

**Describe the bug**
When the progress bar is at the very beginning, the 0% is displayed in the wrong place.

**To Reproduce**
1. Go to Media > Bulk Optimization
2. Launch the optimization
3. See where 0% is displayed

**Expected behavior**
It should be displayed better, under the 

**Screenshots**
1.
image

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

joejoe04:
closes #594 

When optimization starts, the percentage value looks like it's not in the right place, so that has been corrected:

Also, at widths below 782px, the top padding of the rows was reduced and not equal to the bottom padding and it also caused the percentage to overlap with content, so this has been addressed:

Also combined duplicate CSS rules into one.


remyperona (approved):

Mai-Saad:
Can see it fixed on chrome, firefox, responsive view on chrome.
Note:  For Safari on Mac or on ios, can see 0% just on the borderline, can we shift it a little right to be like on chrome? @joejoe04 
Screenshot from 2021-12-13 10-58-35
Screenshot from 2021-12-13 10-57-03

on chrome:
Screenshot from 2021-12-13 10-59-54

joejoe04:
@Mai-Saad Yes sure, I've pushed an update to move the 0% a couple of pixels more to the right, so it should now look okay for all browsers. Thank you!

GeekPress (approved):
QA done, it's working fine

PR #628, opened 2021-11-02
Fix/592 small css glitch in other media section

### LINKED ISSUE #592: Small CSS glitch in Other Media section

**Describe the bug**
There's a small CSS glitch in the media section related to dropdowns

**To Reproduce**
1. `Media > Other Media`
2. See the dropdowns

**Expected behavior**
Better styling of dropdowns

**Screenshots**
image

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

joejoe04:
closes #592 

Fixes issues with the select and filter button display on the "Other Media" page including:

Corrects the select element padding so the text doesn't run into the down arrow icon at narrow widths and also so it matches the height of the filter button:

Fixes filter button padding at width of 782px and below so its height continues to match height of select elements:

Changes the top margin of select elements and filter buttons so they adapt with better spacing at narrower display widths:


remyperona (approved):

Mai-Saad:
Can see the initial problem fixed on chrome, FF, safari. 

Note: `All folders` field width is dependent on selected folder path length i.e for longer path, the width of All folders field increase and corrupt in small screens, Can we fix this? (i.e we can truncate the path of older after a certain length maybe) will this be included in the current PR? @Tabrisrp @joejoe04  what do you think?

Screenshot from 2021-12-13 11-34-55

Screenshot from 2021-12-13 11-38-35

joejoe04:
@Mai-Saad Nice catch on the select element width with long folder names. I’ve added one more commit applying a max-width value to it and it now looks like this at narrow widths:

image

However, that doesn’t help with the option element inside with the folder names:

image

I tried doing some searching, but I’m not sure it’s possible to limit the width of these. If this is an issue we should fix, that may need a separate GH issue, where we could possibly do something like limit the values to a specific number of characters, but that could lead to some confusion possibly.

Interested if anyone else has an opinion on this.

GeekPress (approved):
QA done, everything is good!

PR #631, opened 2021-11-04
Fix/586 remove WP Rocket banner when plugin installed (but inactive)

### LINKED ISSUE #586: WP Rocket banner is also displayed when WP Rocket is installed but deactivated

**Describe the bug**
When WP Rocket is installed but deactivated, we are displaying banner in the Bulk Optimization section. The banner is giving 20% of discount for WP Rocket purchases, not sure if people who already have WP Rocket need it.

**To Reproduce**
1. Install WP Rocket
2. Go to Media > Bulk Optimization
3. See banner

**Expected behavior**
I think we should display banner only if WP Rocket is not installed

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

markonikolic985:
Closes #586 

Hide WP Rocket Banner when plugin is installed but not active.

Changed files:

- [views/page-settings.php](https://github.com/wp-media/imagify-plugin/blob/88e39cf23b75b2e63f253e6e6fcd996e4986935b/views/page-settings.php#L16) to keep `imagify-have-rocket` CSS class so the main column has 100% width
- [views/page-rocket-ad.php](https://github.com/wp-media/imagify-plugin/blob/88e39cf23b75b2e63f253e6e6fcd996e4986935b/views/part-rocket-ad.php#L4-L8) to check if plugin exists (active or not)

### DIFF
--- a/inc/classes/class-imagify-notices.php
+++ b/inc/classes/class-imagify-notices.php
@@ -608,7 +608,9 @@ class Imagify_Notices extends Imagify_Notices_Deprecated {
 			return $display;
 		}
 
-		if ( defined( 'WP_ROCKET_VERSION' ) || self::notice_is_dismissed( 'wp-rocket' ) ) {
+		$plugins = get_plugins();
+
+		if ( isset( $plugins['wp-rocket/wp-rocket.php'] ) || self::notice_is_dismissed( 'wp-rocket' ) ) {
 			return $display;
 		}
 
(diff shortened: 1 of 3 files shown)


Mai-Saad:
Can see the fix on the settings page (currently if WPR is installed or activated, the Banner is not displayed). However, Can't check the Bulk Optimization page as after closing the banner while exploring on the trunk, it didn't appear again even after deleting transients.

**Update:** Tried another account, Banner removed from settings page but not from Bulk Optimization page. @markonikolic985 Can you please check this case?
https://jmp.sh/Un0UdNm

iCaspar:
Hey here @Mai-Saad 
I've added a tweak to hid the banner on the bulk page as well.
Also, if you previously dismissed a banner, you will find in the db -> wp_usermeta table an entry for `_imagify_ignore_ads` that stores the dismissed status. If you remove that, it should reset the banner to the "as new" state.

iCaspar (approved):

GeekPress (approved):
QA done, it's working like expected :)

PR #650, opened 2022-01-03
Fixes #609 conflict with wp cloudflare super page cache

### LINKED ISSUE #609: Conflict with WP Cloudflare Super Page Cache
**Describe the bug**
When [WP Cloudflare Super Page Cache](https://wordpress.org/plugins/wp-cloudflare-page-cache/) plugin is enabled, there is a problem in Bulk Optimization page, as the "Imagif'em all" button does not work. 

The problem is happening because both plugins use the same Sweet Alert script and there is Console error when trying the Bulk Optimization:
Screenshot 2021-07-29 at 15 40 21

**To Reproduce**
1. Activate Imagify and [WP Cloudflare Super Page Cache](https://wordpress.org/plugins/wp-cloudflare-page-cache/)
2. Go to Media > Bulk Optimization
3. Click on "Imagif'em All" button (make sure to open Inspect Element > Console tab)

**Expected behavior**
Popup should appear in Bulk Optimization page and you should be able to start the optimization

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

joejoe04:
closes #609 

When both Imagify and [WP Cloudflare Super Page Cache](https://wordpress.org/plugins/wp-cloudflare-page-cache/) plugins are enabled at the same time, the Bulk Optimization will not start when "Imagif'em All' is pressed. This was due to the fact they both used the same Sweet Alert script and this caused a conflict.

This fixes that conflict by checking if the WP Cloudflare Super Page Cache Sweet Alert script is enqueued, and if so, it dequeues it.

### DIFF
--- a/inc/3rd-party/wp-cloudflare-super-page-cache.php
+++ b/inc/3rd-party/wp-cloudflare-super-page-cache.php
@@ -1,14 +1,14 @@
 <?php
 defined( 'ABSPATH' ) || die( 'Cheatin uh?' );
 
-add_action( 'admin_enqueue_scripts', function($hook_suffix) {
+add_action( 'admin_enqueue_scripts', function( $hook_suffix ) {
 	if (
 		! is_admin()
 		||
 		'media_page_imagify-bulk-optimization' !== $hook_suffix
 		||
 		! class_exists( 'SWCFPC_Backend' )
-	){
+	) {
 		return;
 	}
 
--- a/inc/3rd-party/wp-cloudflare-super-page-cache.php
+++ b/inc/3rd-party/wp-cloudflare-super-page-cache.php
@@ -1,0 +1,53 @@
+<?php
+defined( 'ABSPATH' ) || die( 'Cheatin uh?' );
+
+if ( defined( 'SWCFPC_PLUGIN_PATH' ) ) :
+
+	/**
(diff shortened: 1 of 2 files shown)


inc/3rd-party/wp-cloudflare-super-page-cache.php
```diff
@@ -0,0 +1,16 @@
+<?php
+defined( 'ABSPATH' ) || die( 'Cheatin uh?' );
+
+add_action( 'admin_enqueue_scripts', function($hook_suffix) {
+	if (
+		! is_admin()
+		||
+		'media_page_imagify-bulk-optimization' !== $hook_suffix
```
engahmeds3ed: I believe that @markonikolic985 told me that we need to do that in all imagify admin pages not just only bulk optimization one, right?

iCaspar (approved):

iCaspar:
Ready for @wp-media/qa 
Acceptance criteria: When Cloudflare Super Page Cache and Imagify are both active, there is no JS sweetalert error in console on any of the Imagify plugin admin pages, and bulk works as expected.

GeekPress (approved):
QA done, no conflict anymore.

Issue #549, opened 2021-03-17
Improve the "Generate Missing WebP Versions" Function in the Settings Page


joejoe04:
https://secure.helpscout.net/conversation/1434208379/241822

- Seems there may be a bug where it goes through the process of generating the missing WebP versions, and seems to be doing so successfully, but after the process is complete, the same number of missing WebP images is still reported.
- It's unclear whether the WebP versions are created but the message remains afterwards due to a bug or if the WebP versions are not actually created (even though it seems to be happening successfully during the process).
- It's also possible that backups for these images don't exist, but there's no indication that that is the case

**To Reproduce**
Unable to reproduce this from my end at the moment, but there are a fair amount of users who have this issue where clicking the "Generate Missing WebP Versions" button either doesn't work to create the WebP images, or it does work, but the message that the WebP versions are missing still remains afterward

**Expected behavior**
- Either the WebP versions should be generated and the message about missing WebP versions should be removed, or if the WebP images are unable to be created, there could be some indication of why and for which images this is the case.
- Also, when it lists the number of images with missing WebP versions, it would be very much helpful if it would also list which images it is referring to. 
- Would also be good if there was a way to know if backup images were missing for these images
  - When users have 10s of thousands of images and this isn't working, it's very hard to offer them a good solution because there's no way to know which images it's referring to without searching manually through all images for missing WebP versions and/or original backups.

iCaspar:
Cannot reproduce (yet!).

However, in the case cited in the OP, the problem was specifically that the backed-up originals were missing. In such cases Imagify cannot generate WebP (or for that matter, re-generate any other sizes/compression levels) as the saved originals are required for any optimization.
Perhaps the UI could be more clear about that specific situation. It would require additional check for the existence of a backed-up original at the time of loading any given page of images in the media library, noting that "missing backup" and removing the links for any re-optimization. As it is, I think the behavior shown in the customer's screenshots is the fallback provisionally intended -- the popup says it cannot be done because the backups are missing.

joejoe04:
Doing some additional testing, I think there actually could be at least a couple of issues happening, and I believe I may have found a way to reproduce at least some of them, including the issue where even though backup images do exist, the tool is still refusing to create the WebP versions and also the issue where the "you have N optimized images without WebP versions" is not updated after the process is completed.

[...]

PR #654, opened 2022-01-24
Fix 652: Improve messaging for regenerating missing webp

### LINKED ISSUE #652: Fix "workin on" messaging around the Generate Missing webp feature
Subtask of #549 

From @joejoe04:

To start, I upload 2 images to the Media Library with WebP creation disabled and allow them to optimize. After doing this, I see the settings page say there are 2 images with missing WebP versions. Then I enable WebP creation and delete the backup version of one of the images.

When doing this and clicking "Generate Missing WebP Versions" I see this (even though it says there are 2 missing WebP versions, the tool displays that it is working on 0/1 images.

https://user-images.githubusercontent.com/45705494/145212113-487a228b-a66c-47d1-9aed-a8f6ad6c68d1.png

iCaspar:
This is to accomplish several UI/UX issues regarding the "Regenerate Missing WebP" section of the Settings page.

- Hides the "It seems that you have 1 optimized image without WebP versions." Message after the user clicks the "Regenerate Missing WebP Versions" button. (Allowing the Progress bar to indicate what is happening.)
- Hides the whole section (message and the button) when the regeneration process has completed.
- Updates the count of the message to NOT include images that we can determine beforehand for which we cannot generate webp versions (backup is missing, main filepath is missing, wpdb cannot find filepath for attachment). This clears up the situation where the message is displayed, but then either no webp images are created or fewer images are shown working on the progress bar than were expected from the count in the message.


iCaspar:
Fixes #652 
Related: #549

Issue #490, opened 2020-04-28
Add compatibility with "WP Retina 2x" plugin back


Screenfeed:
WP Retina 2x was supported before we introduce webp support, and dropped because lack of time to adapt it to the new architecture.

NicholasRBowers:
Could also tag this as a feature request - if Imagify (or Imagify + WP Rocket) automatically generated and re-wrote URLs to serve @2x retina files, there wouldn't be a need for WP Retina 2x. I believe Optimole does this automatically.

KingGrizzly:
What's the latest on retina support? Thanks.

Klemart3D:
Any news about this issue? Why Imagify couldn't directly include a Retina image management, so no need to be compatible with a 3rd party plugin 🤔?

tnowak:
The only issue with WP Retina 2x compatibility I see is that Imagify plugin does not create .webp images for:
- full size Retina image (original@2x.png) --> no original@2x.png.webp created
- other sizes such as thumb-150x150@2x.png --> no thumb-150x150@2x.png created

That issue should be very easy to fix.
Sure, I can run a cron to cwebp all *2x.png's in the wp-uploads folder, but it would be much easier if Imagify do that.

piotrbak:
@iCaspar Whenever you have a couple of seconds, could you groom this issue?

iCaspar:
The WP Retina 2x plugin is now called Perfect Images. See https://wordpress.org/plugins/wp-retina-2x/
We can start by grabbing the Imagify-WPR2x compat code as it was just before it was removed: commit #68c489b4
From there, we will need to adapt it to the "new" Imagify architecture, and integrate it with Imagify's webp functionality.
There is quite a bit of code there to review/adapt/refactor and update to make sure we're covering changes that have also occurred in WPR2x in the interim.

Estimate: [L]

saranshj4:
https://secure.helpscout.net/conversation/2670983912/506108?folderId=676697

saranshj4:
https://secure.helpscout.net/conversation/2712929608/513254/

saranshj4:
https://secure.helpscout.net/conversation/2763937495/523504?viewId=8160051

saranshj4:
https://secure.helpscout.net/conversation/3158355104/605782?viewId=8160051

PR #656, opened 2022-02-07
Restore Retina Support (via WP Retina 2x, aka Perfect Images, plugin compat)

iCaspar:
This is to restore basic compatibility with the Perfect Images (formerly "WP Retina 2x"), enabling Imagify to process the retina images generated by Perfect Images, and also re-optimizing images when the Perfect Images "Regenerate  Thumbnails" feature is used.

**NOTE: This PR provides compatibility with the FREE Perfect Images version. Neither the PRO version, nor the new use of Perfect Images Easy IO (EWWW) CDN service are included here.**

**Context:**
Compatibility with WP Retina 2x was removed from Imagify in v 1.9 when the "new" architecture and webp was added. See #490. Since that time, WP Retina 2x became "Perfect Images" and implemented several new features, including a Optimization/CDN partnership with EWWW. On the code side, Perfect Images now uses the WP_Rest API instead of AJAX to initiate functionality.

**Coding Considerations**:
Previously, we attached hooks to WP Retina 2x AJAX hooks, passing our own nonces, etc. With the new Rest API approach, Perfect Images defines each endpoint with its own validation, per the WP endpoint definitions. We hook into the Perfect Images functionality after requests have already been checked and retina processing is already started.

The 3 PI processes we are interested in are: Retina Generation, Regenerate Thumbnails, and remove retina images.

- For both generation processes, Imagify needs to replace the current full-size image (if it has been previously optimized) with the originally uploaded file we have moved to Imagify's backup. This is to insure that the new retinas or regenerated thumbnails will be created from the original image. Then, after the process is completed, we need to put the original uploaded version back into the backup location. This is done in the compatibility class's `restore_originally_uploaded_image()` and `reset_optimized_full_size_image()` methods. (NOTE: similar methods are also used for other compatibilities -- the Regenerate Thumbnails plugin compat, for example. These similar methods should ideally be extracted from the various compatibilities and moved into a "BackupHelper" class of some kind.)

- For the retina generation, we need to track which sizes PI generates. Then, after all the retina sizes are generated, we need to find them and send them to Imagify background process for optimization. We also need to tell Imagify how to handle the @2x filenames in order to generate the related webp versions, when Imagify's "Generate Webp" option is enabled. Collecting the list of image sizes PI is generating retinas for is in `add_retina_size()`. Handling retina filenames and queuing them for optimization is done in `optimize_retina_sizes()`. Finally, after the retina files have been optimized, the optimized image data needs to be stored in the Imagify's data for the attachment/media. That is done in `add_retina_sizes_meta()` (hooked to the job process for each image as it is optimized from the background).

- For the regenerate thumbnails, besides handling the backup and restore of the original, we just need to send the newly regenerated images to Imagify's queue for background processing after PI has done it's thing. That is handled by `reoptimize_regenerated_images()`

- For the deletion of retina images, we need to do 2 things: (1) we need to delete the corresponding webp images that we created during the generation. (PI doesn't know anything about these images, so getting rid of them is on us!) We do that in `remove_retina_webp_size()`. Then (2) we need to remove the data from Imagify's data storage that we saved about the retina sizes -- and their related webp retina sizes. We do that in `remove_imagify_retina_data()`.

_Other "how it works" considerations:_
PI has (Pro) functionality to upload/create a "full size" retina image. The thumbnail retina sizes, as opposed to the full-size, are generated, where possible, from re-sizing the full-size image. So ordinarily any thumbnail sizes that have a dimension more than half the full size will not be generated by PI, the adequate resolution to create the thumbnail not being present. By (pro) uploading an original retina full-size, PI is able to generate all the thumbnail sizes from that 2x original. Provided the pro version uses the same hook when generating retinas from the pro @2x original upload, Imagify should still optimize the retina sizes, as well as optimizing the @2x original upload as it would for any other uploaded image. None of this is verified, however, at this time. It would be great if we could reach out to the PI dev to confirm/add Pro compatibility for this and other pro features.

Also, it's probably out of the picture to be compatible with a direct competitor's Optmimzation/CDN service.

Long term, having native retina support in Imagify would be ideal.


inc/3rd-party/perfect-images/classes/PerfectImages.php
```diff
@@ -0,0 +1,364 @@
...
+	 *
+	 * @hooked imagify_media_files
+	 *
+	 * @param array $sizes The Media's size data.
+	 *
+	 * @return array Sizes data that includes retina sizes.
+	 */
+	public function add_retina_sizes_meta( array $sizes ): array {
+		foreach ( $sizes as $size => $size_data ) {
+			$retina_path = wr2x_get_retina( $size_data['path'] );
```
remyperona, line 166: Maybe a function exists check before trying to use a 3rd party function? We never know

PR #664, opened 2022-05-18
Closes #663 Smart compression in the plugin

### LINKED ISSUE #614: Keep EXIF Data in WebP Images Too if the Option is Enabled in Settings Page
When WebP versions are created, the EXIF data is removed whether or not the option to keep EXIF data is enabled in the Imagify plugin settings page.

It may be helpful if the WebP images adhered to the `Keep all EXIF data from your images` setting as the optimized JPG/PNG images do, so the EXIF data could be preserved in WebP versions as well if that is what the user wants.

**Additional context**
[This is a ticket](https://secure.helpscout.net/conversation/1510195359/263321) that illustrates the problem this may be creating with colors of WebP images sometimes. I've run a few test optimizations to compare resulting color in various cases:

**Here is the Original image**
baum-396-flyer-original

**Here is the Optimized JPG with EXIFs kept (color same as original)**
baum-396-flyer-optimized-with-exif

**Here is the Optimized WebP version (color changed fairly substantially)**
https://jmp.sh/zbXMrmI

**And here is the Optimized JPG when EXIFs are removed (color changed like the WebP version)**
baum-396-flyer-optimized-exif-removed

Given that both the WebP version and the optimized JPG version without EXIF are seeing the same color change (and both are missing the EXIF data), it seems the color change in the WebP version may be likely due to missing EXIF data rather than something related to the WebP file type specifically or the process by which we generate them.

### LINKED ISSUE #659: Keep EXIF Data in thumbnails too after optimization
**Describe the bug**
If a user has the `Keep all EXIF data from your images` option enabled, after optimization, the EXIF data is missing from the thumbnails.

**To Reproduce**
1. Enable the `Keep all EXIF data from your images` option in the Imagify plugin settings page
2. Disable at least one thumbnail size from optimization in the `Files Optimization` section
3. Upload a test image with EXIF data and optimize it
4. You should notice that the original image (or -scaled version if applicable) retains the EXIF data after optimization, but all the thumbnails that were optimized no longer have the EXIF data
5. We can also see the EXIF data remains in the thumbnail size(s) excluded from optimization, so we know the EXIF data is there when WordPress creates the thumbs on upload, but is removed after optimization.

**Expected behavior**
EXIF data should be preserved for all images (original and thumbs) after optimization.

**Additional context**
Here is the user comment that brought this to our attention:
https://wordpress.org/support/topic/metadata-gets-removed-for-thumbnails/

• This is also somewhat related to [Issue #614](https://github.com/wp-media/imagify-plugin/issues/614)

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

### LINKED ISSUE #663: Smart compression in Imagify plugin
Imagify App can handle the smart compression, we need to update plugin to be fully compatible with it.

Epic card can be found here:
https://docs.google.com/document/d/1NcaJc9xcJepH4ErJ4--Y4WNmUEN6Jym1E8ucgWi9ik4/edit?usp=sharing

remyperona:
Update the plugin to use the new smart compression system.

Closes #663
Closes #659
Closes #614

When updating to v2.0:
- Optimization is switched to the smart compression
- Images previously optimized with aggressive or ultra are now showing as smart compression
- Images previously optimized with normal are now showing as lossless compression
- Images optimized with lossless can be re-optimized with smart compression
- Images optimized with smart can be re-optimized with lossless compression
- The optimization level and exif options are removed from the settings
- exif data is always preserved on all images

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Normal compression is a "lossless" optimization. This means there is no loss of 
 
 ### Is the EXIF data of images removed?
 
-By default EXIF data is removed. It is however possible to keep it by enabling the option.
+EXIF data is not removed.
 
 ### Will the original images be deleted?
 
--- a/classes/Context/AbstractContext.php
+++ b/classes/Context/AbstractContext.php
@@ -1,8 +1,6 @@
 <?php
 namespace Imagify\Context;
 
-defined( 'ABSPATH' ) || die( 'Cheatin’ uh?' );
-
 /**
  * Abstract used for contexts.
  *
@@ -16,7 +14,6 @@ abstract class AbstractContext implements ContextInterface {
 	 *
 	 * @var    string
 	 * @since  1.9
-	 * @access protected
 	 * @author Grégory Viguier
 	 */
 	protected $context;
@@ -26,7 +23,6 @@ abstract class AbstractContext implements ContextInterface {
 	 *
 	 * @var    bool
 	 * @since  1.9
-	 * @access protected
 	 * @author Grégory Viguier
 	 */
 	protected $is_network_wide = false;
@@ -39,7 +35,6 @@ abstract class AbstractContext implements ContextInterface {
 	 *     - 'image' to allow only images.
 	 *     - 'not-image' to allow only pdf files.
 	 * @since  1.9
-	 * @access protected
 	 * @see    imagify_get_mime_types()
 	 * @author Grégory Viguier
 	 */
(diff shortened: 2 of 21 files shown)


engahmeds3ed (approved):
Just a question, do we need to check the imagify app itself to remove those options or it's just from the plugin side?

image

remyperona:
No only the plugin

PR #681, opened 2022-08-05
Add API functions and WP CLI commands for bulk optimization

remyperona:
Add 2 new functions to run the bulk optimization and missing WebP generation programatically.

Also add WP CLI commands to do the same.

### DIFF
--- a/classes/Plugin.php
+++ b/classes/Plugin.php
@@ -3,6 +3,8 @@ declare(strict_types=1);
 
 namespace Imagify;
 
+use Imagify\Cli\BulkOptimizeCommand;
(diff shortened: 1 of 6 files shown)


classes/Plugin.php
```diff
@@ -3,6 +3,8 @@
 
 namespace Imagify;
 
+use Imagify\Cli\BulkOptimizeCommand;
```
mostafa-hisham: Cli should be CLI
```suggestion
use Imagify\CLI\BulkOptimizeCommand;
```

mostafa-hisham (changes requested):

mostafa-hisham (approved):

PR #683, opened 2022-08-08
Initialize Bulk instance and update methods signatures

remyperona:
Initialize the bulk instance to register the hooks.

Update methods signature to receive individual parameters instead of an array.


classes/Bulk/CustomFolders.php
```diff
@@ -67,7 +68,7 @@ public function get_unoptimized_media_ids( $optimization_level ) {
 
 		// We need to output file URLs.
```
mostafa-hisham: this comment is confusing should be removed

mostafa-hisham (approved):

PR #687, opened 2022-09-21
Closes #686 Re-introduce lossless compression

### LINKED ISSUE #686: Re-introduce Lossless Compression option
## What are we proposing to do? 
With this update, we want to add an option to select lossless compression and not force the usage of Smart compression by default.

All details: https://docs.google.com/document/d/1pVuDP87-L-O1uH_oV2eqmfJDBrCLZ9drtFWqrx8WZoE/edit?usp=sharing

remyperona:
Re-introduce lossless compression option, with a new setting in the settings page. Also re-add the level selection on the bulk optimization page.

### DIFF
--- a/classes/Job/MediaOptimization.php
+++ b/classes/Job/MediaOptimization.php
@@ -1,56 +1,38 @@
 <?php
 namespace Imagify\Job;
 
-use Imagify\Media\MediaInterface;
-use Imagify\Optimization\File;
 use Imagify\Optimization\Process\ProcessInterface;
-
-defined( 'ABSPATH' ) || die( 'Cheatin’ uh?' );
+use Imagify\Traits\InstanceGetterTrait;
+use WP_Error;
 
 /**
  * Job class for media optimization.
  *
- * @since  1.9
- * @author Grégory Viguier
+ * @since 1.9
  */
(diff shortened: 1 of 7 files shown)


mostafa-hisham (approved):

GeekPress:
I got the exact same optimization on images by using the Smart or Lossless compression.

I tried the same images on the app, and the results are different between Lossless and Smart compression. The lossless compression doesn't seem to be taken into consideration by the plugin.

On the Lossless option, please add an extra break line between the sentences.

remyperona:
Can you send me the image(s) you tested with, and also where did you test (on library page or on bulk optimization).

I tested some images on my side but they're optimized as expected, when comparing with the app results.

GeekPress:
Ok never mind, I had this result because WebP option was turned on.

To avoid this confusion, let's remove the re-optimization link in the media list and the compression mode in the Bulk Optimization to avoid confusion.

It seems we have just one compression mode on WebP optimization.

GeekPress:
Never mind about the WebP compression. It's a bug from the API. We must have a normal compression level too.

We can keep the UI as it is. 

The QA is finished and validated.

PR #694, opened 2022-10-18
Fixes #521 Correctly resize the WebP full image

### LINKED ISSUE #521: WebP is not resized to defined size
When using image resizing in Imagify, original image is resized to defined value, but WebP stays in original size (width/height).

**To Reproduce**

1. Go to Imagify Settings and make sure Resizing and WebP creation options are active
2. Add new Media (or optimize existing) - make sure it is larger in width (so it get resized)
3. Check original image and WebP dimensions. WebP will not be resized.

**Expected behavior**

WebP should be resized as same as the original format image.

**Screenshots**

Screen Shot 2020-10-09 at 11 16 25 AM

remyperona:
Use the file dimensions data to check for the resize needs instead of using the database data

Fixes #521

### DIFF
--- a/classes/Optimization/Process/AbstractProcess.php
+++ b/classes/Optimization/Process/AbstractProcess.php
@@ -1174,13 +1174,7 @@ abstract class AbstractProcess implements ProcessInterface {
 			];
 		}
 
-		$media      = $this->get_media();
-		$dimensions = $media->get_dimensions();
-
-		if ( ! $dimensions['width'] ) {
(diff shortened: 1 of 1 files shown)


mostafa-hisham (approved):

GeekPress:
it's working as expected. WebP original image is correctly resized.

GeekPress:
I moved back to ready for QA as I have maybe found a regression. I need further tests to confirm.

GeekPress:
Ok, after further tests, what I found is an unknown bug, not a regression.

PR #716, opened 2023-03-22
Added a way fix filter issue on type from media

### LINKED ISSUE #670: Media Library Filter does not work in combination with file type
When on Media Library (List VIew), there is possibility to filter files by type, date and optimization status (this one is coming from Imagify).

However, if you choose specific file type (images, documents...) and specific optimization status, the filter will not work - all file types are still displayed.

This works with date filter, but not with file type filter.

**To Reproduce**
1. Go to Media Library
2. Choose List View
3. Choose Documents or Images from the first filter
4. Choose Optimized/Unoptimized/Errors from the third filter
5. Still all file types will show up
6. If you change only one filter, it will work

**Screenshots**
Screenshot 2022-05-24 at 13 46 54

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
Fixes #670 
Prevent override from the `post_mime_type` offset when it is already defined to make filters work correctly.

### DIFF
--- a/inc/admin/upload.php
+++ b/inc/admin/upload.php
@@ -123,7 +123,9 @@ function _imagify_sort_attachments_by_status( $vars ) {
 		),
 	) );
 
-	$vars['post_mime_type'] = imagify_get_mime_types();
+	if ( ! key_exists( 'post_mime_type', $vars ) ) {
+		$vars['post_mime_type'] = imagify_get_mime_types();
+	}
 
 	return $vars;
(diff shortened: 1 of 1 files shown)


remyperona (approved):

Mai-Saad (approved):
Working as expected and the filter was applied correctly

PR #717, opened 2023-03-24
Refactored user, added grunt to package.json and added out of quota popup

### LINKED ISSUE #714: Improve information related to out of quota on the bulk
Once the bulk optimization is launched and all the quota is consumed, we don't directly display the information about the quota. Instead it looks like the bulk is continuing.

That's for example how this is displayed on the bulk (the user has consumed all his quota) : 
image

Instead we should directly use the popin we were using before to warn the user about the quota and provide him a solution.

CrochetFeve0251:
Fixes #714 
Add a popup out of quota that displays also on each time we load the bulk page when the quota is out.
Also this PR refactor User class, add grunt into `package.json` to simplify usage from js assets and finally add `node_modules` folder to phpcs exclusions.


remyperona (approved):

vmanthos (approved):
Working as expected! :+1:

PR #730, opened 2023-05-24
Fixes #729 Escape error message before printing

### LINKED ISSUE #729: Imagify audit enhancements

**Describe the bug**
We received an audit for Imagify plugin, there's one small enhancement to be made.

**Additional context**
The details are present on the 3rd page of this document:
https://drive.google.com/file/d/1BczI8eSF_TDxRGM9712LzQ5i8jkLGajs/view?usp=sharing

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
Following the audit, use `esc_html()` to escape the notice message before printing it.

Fixes #729

### DIFF
--- a/views/notice-temporary.php
+++ b/views/notice-temporary.php
@@ -15,7 +15,7 @@ foreach ( $notices as $type => $type_notices ) {
 	?>
 	<div class="<?php echo $type; ?> settings-error notice is-dismissible">
 		<?php foreach ( $type_notices as $details ) { ?>
-			<p><strong><?php echo $details['message']; ?></strong></p>
+			<p><strong><?php echo esc_html( $details['message'] ); ?></strong></p>
 		<?php } ?>
 	</div>
 	<?php
--- a/views/notice-temporary.php
+++ b/views/notice-temporary.php
@@ -15,7 +15,7 @@ foreach ( $notices as $type => $type_notices ) {
 	?>
 	<div class="<?php echo $type; ?> settings-error notice is-dismissible">
 		<?php foreach ( $type_notices as $details ) { ?>
-			<p><strong><?php echo wp_kses( $details['message'], [ 'code' ] ); ?></strong></p>
+			<p><strong><?php echo wp_kses( $details['message'], [ 'code' => [] ] ); ?></strong></p>
 		<?php } ?>
 	</div>
 	<?php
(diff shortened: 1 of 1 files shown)


nicomollet (approved):

vmanthos:
@Tabrisrp We have a regression in the way messages are displayed, e.g.:
image

with `trunk` is:
image

The HTML in the first case (regression) is:
```HTML
<div class="error settings-error notice is-dismissible">
					<p><strong>Imagify could not insert contents into the .htaccess file: The &lt;code&gt;.htaccess&lt;/code&gt; file is not writable.</strong></p>
			<button type="button" class="notice-dismiss"><span class="screen-reader-text">Dismiss this notice.</span></button></div>
```
while in the correct one:
```HTML
<div class="error settings-error notice is-dismissible">
					<p><strong>Imagify could not insert contents into the .htaccess file: The <code>.htaccess</code> file is not writable.</strong></p>
			<button type="button" class="notice-dismiss"><span class="screen-reader-text">Dismiss this notice.</span></button></div>
```
**Diff:** https://www.diffchecker.com/sEJ7iPeJ/

**To reproduce**

1. Change the permissions of the `.htaccess` file, e.g. `sudo chmod 444 .htaccess`.
2. Install Imagify and enable WebP.
3. Check the admin notice after saving the settings in step 2.

Can you please look into this? :pray:

remyperona:
I updated the escaping, I think it should be good now.

vmanthos:
`<code></code>` is now removed and the notice is displayed like this:
image

instead of 
image

@wp-media/productimagify Are we good with this?

remyperona:
Ok the code tag should be preserved now

vmanthos:
@Tabrisrp The tag is preserved now, thank you!

Is it possible that more tags will be used in the specific template, and therefore will need to be preserved with `wp_kses()`?
https://github.com/wp-media/imagify-plugin/blob/0abba41828b05dc5602f300c5306d9e7c0a33665/views/notice-temporary.php#L18

Temporary notices are added using the `add_site_temporary_notice()` method:
https://github.com/wp-media/imagify-plugin/blob/1442452affd32be5e7aaebb31611d211fe4a405a/classes/Notices/Notices.php#L775-L794

Apart from the scenario with the `.htaccess` that I already checked, the places where that's used through [imagify_maybe_redirect()](https://github.com/wp-media/imagify-plugin/blob/1442452affd32be5e7aaebb31611d211fe4a405a/inc/functions/admin.php#L327-L367) is in `inc/classes/class-imagify-admin-ajax-post.php`

Most of the time, that will have the message of `WP_Error`. Does that contain tags?

remyperona:
As far as I can see, it's the only tag used, but it's hard to find things in Imagify

vmanthos (approved):
Thank you, @Tabrisrp! 

Then this is good to go! :+1:

PR #738, opened 2023-08-01
WordPress 6.3 Image Compatibility Fix

michaelbourne:
Update Display.php to reflect WordPress 6.3 addition of `fetchpriority` and `aspect-ratio` on images. Also added in `decoding` to the list of ignored attributes.


michaelbourne (commented):
Linted

PR #751, opened 2023-10-24
Closes 2554: Prevent having a webp bigger than the original

CrochetFeve0251:
## Description
Fixes #2554

Display the message when we display a original format instead of the webp one.

## Type of change

*Please delete options that are not relevant.*

- New feature (non-breaking change which adds functionality).

## Is the solution different from the one proposed during the grooming?

No.

# Checklists

## Generic development checklist

- [ ] My code follows the style guidelines of this project, with adapted comments and without new warnings.
- [ ] I have added unit and integration tests that prove my fix is effective or that my feature works.
- [ ] The CI passes locally with my changes (including unit tests, integration tests, linter).
- [ ] Any dependent changes have been merged and published in downstream modules.
- [ ] If applicable, I have made corresponding changes to the documentation. *Provide a link to the documentation.*

## Test summary

- [ ] I triggered all changed lines of code at least once without new errors/warnings/notices.
- [ ] I validated all Acceptance Criteria of the related issues. (If applicable, provide proof).
- [ ] I validated all test plan the QA Review asked me to.

*If not, detail what you could not test.*

*Please describe any additional tests you performed.*

### DIFF
--- a/inc/classes/class-imagify-files-list-table.php
+++ b/inc/classes/class-imagify-files-list-table.php
@@ -612,6 +612,10 @@ class Imagify_Files_List_Table extends WP_List_Table {
 						<span class="data"><?php esc_html_e( 'WebP generated:', 'imagify' ); ?></span>
 						<strong class="data-value"><?php echo esc_html( $has_webp ); ?></strong>
 					</li>
+					<li class="imagify-data-item">
+						<span class="data"><?php esc_html_e( 'WebP generated:', 'imagify' ); ?></span>
+						<strong class="data-value"><?php echo esc_html( $has_webp ); ?></strong>
+					</li>
 					<?php
 				}
 			}
--- a/classes/Optimization/Data/AbstractData.php
+++ b/classes/Optimization/Data/AbstractData.php
@@ -25,6 +25,7 @@ abstract class AbstractData implements DataInterface {
 	 */
 	protected $default_optimization_data = [
 		'status' => '',
+		'message' => '',
 		'level'  => false,
 		'sizes'  => [],
 		'stats'  => [
--- a/classes/Optimization/Data/CustomFolders.php
+++ b/classes/Optimization/Data/CustomFolders.php
@@ -159,6 +159,10 @@ class CustomFolders extends AbstractData {
 				$old_data['hash'] = md5_file( $file_path );
 			}
 
+			if ( key_exists( 'message', $data ) ) {
+				$old_data['message'] = $data['message'];
+			}
+
 			if ( ! $data['success'] ) {
 				/**
 				 * Error.
--- a/classes/Optimization/Data/WP.php
+++ b/classes/Optimization/Data/WP.php
@@ -93,7 +93,12 @@ class WP extends AbstractData {
 			'original_size'  => 0,
 			'optimized_size' => 0,
 			'percent'        => 0,
+			'message'        => '',
 		], $old_data['stats'] );
+
+		if ( key_exists( 'message', $data ) ) {
+			$old_data['message'] = $data['message'];
+		}
 
 		if ( ! $data['success'] ) {
 			/**
--- a/classes/Optimization/File.php
+++ b/classes/Optimization/File.php
@@ -530,6 +530,10 @@ class File {
 			return new \WP_Error( 'temp_file_not_found', $temp_file->get_error_message() );
 		}
 
+		if(property_exists($response, 'message' )) {
+			$args['convert'] = '';
+		}
+
 		if ( 'webp' === $args['convert'] ) {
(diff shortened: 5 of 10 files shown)


inc/classes/class-imagify-files-list-table.php
```diff
@@ -612,6 +612,10 @@ public function column_optimization( $item ) {
 						<span class="data"><?php esc_html_e( 'WebP generated:', 'imagify' ); ?></span>
 						<strong class="data-value"><?php echo esc_html( $has_webp ); ?></strong>
 					</li>
+					<li class="imagify-data-item">
+						<span class="data"><?php esc_html_e( 'WebP generated:', 'imagify' ); ?></span>
+						<strong class="data-value"><?php echo esc_html( $has_webp ); ?></strong>
+					</li>
```
remyperona, line 618: This list item is the same as the list item right above?

remyperona (changes requested):
I'm not understanding how the changes are preventing the WebP bigger than original?

Also, the referenced issue is not correct.

inc/classes/class-imagify-files-list-table.php
CrochetFeve0251: The idea of this part is to give a message when the webp is not generated cause we keep the original with optimizations.

Also this PR is in two parts and the actual logic to prevent this is on the python part: https://github.com/wp-media/imagify/pull/2567

remyperona:
We also have this PR related https://github.com/wp-media/imagify-plugin/pull/706

remyperona (approved):

Mai-Saad:
@CrochetFeve0251 Thanks for the update. Please find exploratory test notes below
1- The served image at the FE has size > the optimized size (for this image WebP isnot generated, however it's served, shouldn't we serve png in this case? even if the option to display WebP image on the website is enabled @wp-media/productimagify ) test page https://imagify.rocketlabsqa.ovh/test-webp/
Screenshot from 2023-11-08 09-23-55
Screenshot from 2023-11-08 09-22-06
2- WebP generated is always no although it was generated (@wp-media/productimagify what shall be the text for convert when the image is converted to WebP? do we really need this field? /cc @MathieuLamiot )
Screenshot from 2023-11-08 09-38-48
**on trunk**
Screenshot from 2023-11-08 09-51-55

remyperona (approved):

MathieuLamiot:
@Mai-Saad, thanks for the detailed report.
About the 2nd point: 
- If the WebP is generated (because it is smaller), then WebP Generated should be set to "Yes".
- If the WebP is not generated (because it is bigger, or for any other reasons actually), then WebP generated should be set to "No".
I think it should be fixed as part of this issue @CrochetFeve0251 

I do agree it seems a bit redundant with the message above saying WebP is not generated because it is bigger. It could be a small simplification, but let's see what @wp-media/productimagify thinks about this. We might not have all the cases in minds. In any case, that would be outside the scope of the issue..

Mai-Saad:
@CrochetFeve0251 Thanks for the update. @MathieuLamiot Thanks for the feedback.
Currently:
- Specific image has no WebP generated while status is yes
Screenshot from 2023-11-09 18-04-06

- An image which isnot converted, is served with a bigger size (tested on chrome ubuntu) (note: this image wasnot compressed somehow)

Screenshot from 2023-11-09 18-06-32

Screenshot from 2023-11-09 18-05-46

Mai-Saad:
@CrochetFeve0251 Thanks for the update. Working as the following now. If agreed by @wp-media/productimagify then we can merge. /cc @MathieuLamiot 

1. Original image is non webp

- If generated WebP is < original image => no convert msg, WebP generated status is yes and served on FE if the serve option is enabled
- If generated WebP is > original image => convert msg `Webp is less performant than original`, WebP generated status is either No (if no webp file added to upload folder) or partial (if the webP file added to upload folder), image.extension is served on FE (as same as that in upload folder without extension nor dimensions and it is <= original size) if the serve webP option on/off

2- Original image is webP

- no convert msg, WebP generated status is No, The optimized image is served (approximately as same as that in upload folder )

3- Already on trunk, origional/ optimized size is not matching that in the upload folder https://wp-media.slack.com/archives/CU0F6EGQ1/p1699598198861799

4- Already on trunk, while display image in webp isnot checked, the used size at FE is that of webp but without adding webp extension https://wp-media.slack.com/archives/CU0F6EGQ1/p1699618969452739

DahmaniAdame:
Looks good. The only change is on the wording to make it clearer about why the origin image will be served instead: 

Change: 

```Webp is less performant than original```

To:

```WebP file is larger than the original image```

Mai-Saad:
Screenshot from 2023-11-13 15-58-06

PR #807, opened 2024-02-20
Closes #790, #792, #803: Delete/Generate next-gen in media not working

jeawhanlee:
# Description
Fixes the delete/generate next-gen link not working.

Fixes #790, #792, #803 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue).

# Checklists

## Feature validation

- [x] I validated all the Acceptance Criteria. If possible, provide sreenshots or videos.
- [x] I triggered all changed lines of code at least once without new errors/warnings/notices.

## Code style
- [x] I wrote self-explanatory code about what it does.
- [x] I wrote comments to explain why it does it.
- [x] I named variables and functions explicitely.
- [x] I did not introduce unecessary complexity.

PR #811, opened 2024-02-23
Closes #796: Auto AVIF Convertion on settings save

### LINKED ISSUE #796: AVIF images arenot automatically generated for existing images after enabling the option
- Made sure you’re on the latest version => Avif commit: https://github.com/wp-media/imagify-plugin/commit/b396de27cda71d9b16e311a0f55f482bebf13380

**Describe the bug**
When enabling AVIF, it isnot automatically generate AVIF for existing images in the media library

**To Reproduce**
1. Fresh activation of imagify
2. Upload a couple of images
3. Enable create AVIF option and save 
4. Check FTP for images in 2 => AVIF not generated

**Expected behavior**
AVIF version is generated and quota is affected 

**Additional context**
- related discussion https://wp-media.slack.com/archives/C069W48E47N/p1706878004622759?thread_ts=1706877044.964199&cid=C069W48E47N

Miraeld:
# Description

Fixes #796 

### User documentation

When uploading images to wordpress, when the latest is over a certain dimension, WordPress will resize it and named it `-scaled`. Originally, we were doing a backup of the Original image, while on the bulk optimization it's trying to reach the `-scaled` version as WordPress takes it as default. 
So we've added the `-scaled` version to be backed-up.

### Technical documentation

Before trying to save the `-scaled` version of the image in backup, we check if it exists. Same with the deletion of the image.

## Type of change

# Checklists

## Feature validation

## Code style

## Observability
- [x] I handled errors when needed.


Miraeld:
@Mai-Saad , 
The problem we encountered with big images like this was the fact that WordPress is resizing the image and name it `-scaled`, so for example, for an image `bigimage.jpg` it would become `bigimage-scaled.jpg`. In the DB, the `-scaled` version is saved.

So while trying to get back the IDs of unoptimized medias while running the bulk optimization, it wouldn't keep it as the backup file does not exists. 

To counter that issue, I've added the `scaled` version to the backup, as that's the main one Wordpress is using, and the bulk optimization will work without trouble. 

I've also taken care of the deletion of the image, when we delete the image, it will delete both backup, the original and the `scaled` version.

Khadreal (approved):

Mai-Saad:
@Miraeld Thanks for the PR, now AVIF is created for large image with the following notes.
1- if the image have webp before enable avif, we are deleting the webp version before generating the avif (this is not happening with small image, here we add avif and keep webp there). @piotrbak shouldn't it be same as with small image?
2- if we uploaded image(s) before activating imagify. Those images won't be automatically processed after activate imagify and enable AVIF. @piotrbak are we ok here or we need separate GH?
3- if internet went down while optimization is in-progress then once  back, it won't be resumed automatically. (probably same on trunk, needs further investigation and may need separate GH)

piotrbak:
@Mai-Saad @Miraeld 1st and 2nd point should be fixed.

Miraeld:
Hello hello, 

@Mai-Saad @piotrbak , 
for the 1st point, it's been fixed in the before last commit. 
2nd point is fixed by the last commit.

classes/Optimization/Process/AbstractProcess.php
```diff
@@ -1447,7 +1454,7 @@ protected function delete_nextgen_file( $file_path ) {
 		$next_gen_file = new File( $file_path );
 
 		// Delete next-gen images.
-		foreach ( $this->extensions as $extension ) {
+		foreach ( imagify_nextgen_images_formats() as $extension ) {
```
remyperona: This needs to stop being changed, on a previous PR we used the `$extensions` array because we want to remove all next-gen files when deleting the file, whatever is enabled at that time.

Why is it changed back to the function that will only return formats that are currently enabled?

Miraeld: According to @Mai-Saad , her first feedback, she said we were deleting the webp when we activate AVIF, and AVIF would be generated afterwards. Therefore, we would need to change the behaviour to not delete the webp when we generate the avif. 

When looking at why it gets deleted, when an optimization is started, we get to a point where https://github.com/wp-media/imagify-plugin/blob/a3e791d057ed0acff4f2af03a943c322a3d3465c/classes/Optimization/Process/AbstractProcess.php#L874 this function is called. 
and from this function we get : 
https://github.com/wp-media/imagify-plugin/blob/a3e791d057ed0acff4f2af03a943c322a3d3465c/classes/Optimization/Process/AbstractProcess.php#L1449


If we use `$this->extensions` we deletes every format that are included in this array:
https://github.com/wp-media/imagify-plugin/blob/a3e791d057ed0acff4f2af03a943c322a3d3465c/classes/Optimization/Process/AbstractProcess.php#L113-L116

Miraeld: In my point of view, `webp` files should be deleted when we are enabling `avif` and the opposite as well. Otherwise we would use disk space for nothing with one format. But that's another story. 

I did not think that this function was used to delete the file..

jeawhanlee: Screenshot 2024-02-27 at 10 19 34

Based on @Mai-Saad point [here](https://github.com/wp-media/imagify-plugin/pull/807#issuecomment-1959236989) we adjusted that part to delete everything for that scenario when the link in the image above is clicked but this looks like a different scenario where the `delete_nextgen_files` is being used as well, @Miraeld I think we can adjust the method arguments to cater for both scenario.

https://github.com/wp-media/imagify-plugin/blob/a62dc23cb22b6c9f1ba052f17f4d670728d160e4/classes/Optimization/Process/AbstractProcess.php#L1386

We can add another argument ($all_next_gen) and keep it false by default. we need to parse this value down to the base method that is being called [here](https://github.com/wp-media/imagify-plugin/blob/a62dc23cb22b6c9f1ba052f17f4d670728d160e4/classes/Optimization/Process/AbstractProcess.php#L1411).
We can now call `imagify_nextgen_images_formats()` when the value is false and use $this->extensions instead when it is true.

We need to remember to set the new arg to true here https://github.com/wp-media/imagify-plugin/blob/a62dc23cb22b6c9f1ba052f17f4d670728d160e4/inc/classes/class-imagify-admin-ajax-post.php#L236

Then we can update all call occurrence of this method alongside interface.

cc: @Miraeld @Tabrisrp What do you think?

Mai-Saad:
@Miraeld Thanks for the update, compared to trunk and feature/avif branch, we will have regression here:
1- Create avif is enable
2- upload image => image have avif created
3- disable avif
4- permanently delete the image from media library => avif won't be removed from upload folder (same will occur if webp image was created while avif is enabled and we try to permanently delete the image)
can you please check?

classes/Optimization/Process/AbstractProcess.php
Miraeld: I was thinking about adding a param to the function yea, let me take a look at the code and make a commit :)

piotrbak:
@Miraeld Should it be moved to the QA again? Or it needs another CR?

Miraeld:
Umm, 
I think it doesn't need another CR as I just applied what @jeawhanlee advised. 
I can move it back to QA

Mai-Saad:
@Miraeld Thanks for the update.
- The regression mentioned [here](https://github.com/wp-media/imagify-plugin/pull/811#issuecomment-1966355663) is now fixed
- Remaining that now, for large image uploaded before enable AVIF, we reoptimize the image i.e update timestamp for optimized images => we shouldn't reoptimize the image  but just add avif, no? /cc @piotrbak  (Note: after enable avif, if we refresh media and click generate link before background process start, it will work correctly i.e just generating avif so may be we can compare the generate code to the code on PR to catch the root cause)
**update: remaining issue(s) already on main branch will open separate GH about it**

Issue #760, opened 2023-11-21
Generate Missing WebP progress bar can be -ve in specific case


Mai-Saad:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version => 2.1.3
- Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
Generation of missing webp can be -ve if we uploaded images while generation is loading

**To Reproduce**
1. Install and activate imagify
2. click generate missing webp images btn
3. make sure that `Auto-Optimize images on upload` is checked
4. upload some images
5. notice the progress bar of generating missing webP images

**Expected behavior**
The progress bar is +ve, showing progress and moving

**Screenshots**
![Screenshot from 2023-11-21 12-10-55](https://github.com/wp-media/imagify-plugin/assets/76941962/bfe8b1d8-99fa-4179-b76b-ab28afd61441)

**Additional context**
- before uploading images, the progress bar kept loading without any updates i.e 0/23 as if it isnot working

**Acceptance Criteria (for WP Media team use only)**
Clear instructions for developers, to be added before the grooming

Mai-Saad:
for test: when fixed, we need to check this https://github.com/wp-media/imagify-plugin/pull/820#issuecomment-1973004389

PR #820, opened 2024-02-26
Fix data for optimized medias without next-gen versions

remyperona:
# Description

This PR updates the code related to storing and using the data for optimized medias without next-gen versions (depending on option selected, either WebP or AVIF).

### Technical documentation

The logic remains the same, but makes use of the new method to get the list of optimized medias without next-gen.

It also removes no longer used methods that were only accounting for WebP.

## Type of change

# Checklists

## Feature validation

## Code style


classes/Bulk/AbstractBulk.php
```diff
@@ -105,13 +105,19 @@ function_exists( 'set_time_limit' )
...
-	 * Tell if there are optimized media without WebP versions.
+	 * Tell if there are optimized media without next-gen versions.
 	 *
 	 * @since 1.9
 	 *
 	 * @return int The number of media.
 	 */
-	public function has_optimized_media_without_webp() {
-		return count( $this->get_optimized_media_ids_without_webp()['ids'] );
+	public function has_optimized_media_without_nextgen() {
```
jeawhanlee, line 114: This method seems to cater only for the enabled option, don't we need to handle any next-gen version irrespective of the enabled one?

classes/Stats/OptimizedMediaWithoutNextGen.php
```diff
@@ -1,44 +1,37 @@
...
-	/**
-	 * Launch hooks.
-	 *
-	 * @since 1.9
-	 */
-	public function init() {
-		add_action( 'imagify_after_optimize',      [ $this, 'maybe_clear_cache_after_optimization' ], 10, 2 );
-		add_action( 'imagify_after_restore_media', [ $this, 'maybe_clear_cache_after_restoration' ], 10, 4 );
-		add_action( 'imagify_delete_media',        [ $this, 'maybe_clear_cache_on_deletion' ] );
+	public static function get_subscribed_events() {
```
jeawhanlee, line 28: Codacy complains about no doc-block for this

classes/Bulk/AbstractBulk.php
remyperona: No because in that case we want to check based on the currently enabled option

Mai-Saad:
@Tabrisrp Thanks for the PR. Is there any snippet we can use to control / overwrite next-gen status of image(s) so we can easily validate possible scenarios here?

remyperona:
I added a method to automatically clear the counter transient when enabling/disabling `convert_to_avif`

Mai-Saad:
@Tabrisrp Thanks for the PR. 
- The generate count is updated when activate/deactivate avif 
- Validated if avif only or webp only is enabled 
- Out of scope (not handled yet and need snippet to facilitate test) : when avif is enabled + we enable webP using filter => we can have separate GH about that
- We can have -ve missing images in the following case (can have separate GH or just add comment [here](https://github.com/wp-media/imagify-plugin/issues/760)) 

1. avif is disabled
2. upload some images (webP is created)
3. enable webP using filter
4. enable avif

note: without step 3, it will work fine

PR #824, opened 2024-02-27
Apply nginx rules for next gen images

### LINKED ISSUE #821: R&D for NGINX AVIF/WEBP rewrite rules
We need to do a research related to AVIF/WEBP NGINX rewrite rules. The objective is for them to work on:
1. Local NGINX environments
2. Our WordOps environment

The AC:
1. When browser accepts AVIF and WebP, the order of displaying should be 'AVIF > WEBP > Optimized Image'
2. If browser doesn't accept AVIF, we should display: 'WEBP > Optimized Image'
3. If browser doesn't accept WEBP, we should display 'Optimized image'
4. It should be possible by adding the rewwrite rules only to the `server` directive

There is no possibility that the browser accepts AVIF but no WEBP:
https://caniuse.com/?search=webp
https://caniuse.com/?search=avif

wordpressfan:
# Description

Fixes #821

### User documentation

Here we are playing with rewrite rules so users need to reload their nginx after upload to get those new rules being applied.

### Technical documentation

Here I used the webp class to write the next gen handling rewrite rules so we don't need to run any code with plugin update to remove the old rules and insert the new ones, as I think those webp rules were already there in previous versions.

Also I kept avif rewrite rules empty to unify the code structure and keep it as currently used, but it's not added to the config file at all because it's empty.

If you think this would be confusing having the comment `Imagify: rewrite rules for webp` let me know and I can check how to do this upgrade routine here in imagify.

## Type of change
*Delete options that are not relevant.*

## New dependencies

## Risks
*List possible performance & security issues or risks, and explain how they have been mitigated.*

As mentioned above we have two risks:
1. We need to tell users to restart their nginx once our rules are changed, maybe we can use admin notices or add this to our documentation in a clear way.
2. We need to check if the comment in config file is confusing or it;s OK to have `Imagify: rewrite rules for webp` but it's handling webp and avif.

What do u think @piotrbak ?

# Checklists

## Feature validation

## Documentation

- [ ] I prepared the user documentation for the feature/enhancement and shared it in the PR or the GitHub issue.
- [ ] I prepared the technical documentation if needed, and shared it in the PR or the GitHub issue.

## Code style

## Observability
- [ ] I handled errors when needed.
- [ ] I wrote user-facing messages that are understandable and provide actionable feedbacks.
- [ ] I prepared ways to observe the implemented system (logs, data, etc.).

## Risks
- [x] I explicitely mentioned performance risks in the PR.
- [ ] I explicitely mentioned security risks in the PR.

### DIFF
--- a/classes/Avif/RewriteRules/Nginx.php
+++ b/classes/Avif/RewriteRules/Nginx.php
@@ -25,24 +25,6 @@ class Nginx extends AbstractNginxDirConfFile {
 	 * @return string
 	 */
 	protected function get_raw_new_contents() {
-		$extensions = $this->get_extensions_pattern();
-		$extensions = str_replace( '|avif', '', $extensions );
-		$home_root  = wp_parse_url( home_url( '/' ) );
-		$home_root  = $home_root['path'];
-
-		return trim( '
-location ~* ^(' . $home_root . '.+)\.(' . $extensions . ')$ {
-	add_header Vary Accept;
-
-	if ($http_accept ~* "avif"){
-		set $imavif A;
-	}
-	if (-f $request_filename.avif) {
-		set $imavif  "${imavif}B";
-	}
-	if ($imavif = AB) {
-		rewrite ^(.*) $1.avif;
-	}
-}' );
+		return '';
 	}
 }
--- a/classes/Webp/RewriteRules/Nginx.php
+++ b/classes/Webp/RewriteRules/Nginx.php
@@ -34,9 +34,9 @@ class Nginx extends AbstractNginxDirConfFile {
 
 		return trim( '
 location ~* ^(' . $home_root . '.+)\.(' . $extensions . ')$ {
-	add_header Vary Accept;
+    add_header Vary Accept;
 
-	set $canavif 1;
+    set $canavif 1;
 
     if ($http_accept !~* "avif"){
         set $canavif 0;
--- a/classes/Webp/RewriteRules/Nginx.php
+++ b/classes/Webp/RewriteRules/Nginx.php
@@ -29,7 +29,6 @@ class Nginx extends AbstractNginxDirConfFile {
 	 */
 	protected function get_raw_new_contents() {
 		$extensions = $this->get_extensions_pattern();
-		$extensions = str_replace( '|webp', '', $extensions );
 		$home_root  = wp_parse_url( home_url( '/' ) );
 		$home_root  = $home_root['path'];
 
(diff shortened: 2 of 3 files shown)


classes/Webp/RewriteRules/Nginx.php
jeawhanlee, line 1: Shouldn't we update the value of the class const `TAG_NAME` to be generic?
Don't we also need to move it out of the Webp namespace?

classes/Avif/RewriteRules/Nginx.php
jeawhanlee, line 1: Should we still keep this class?

inc/functions/attachments.php
```diff
@@ -19,6 +19,7 @@ function imagify_get_mime_types( $type = null ) {
 			'png'          => 'image/png',
 			'gif'          => 'image/gif',
 			'webp'         => 'image/webp',
+			'avif'         => 'image/avif',
```
remyperona: this should not be added, it's causing bugs in some cases (we already reverted it in a previous PR)

classes/Webp/RewriteRules/Nginx.php
remyperona: I agree it feels weird to do it like this

remyperona (changes requested):

classes/Webp/RewriteRules/Nginx.php
wordpressfan: I mentioned that in the PR description (risks part) but let me explain this in details, the idea here is, currently our customers have old webp rules and if I understand this part well, we are detecting those rules by this tag name, so if I changed this tag name without adding an update routine to completely remove the old rules, we will end up having two rules, something like the following:

```
# BEGIN Imagify: rewrite rules for webp
OLD RULES
# END Imagify: rewrite rules for webp

# BEGIN Imagify: rewrite rules for nextgen
NEW RULES
# END Imagify: rewrite rules for nextgen
```

but if we kept it as webp we will not need this update routine.

If u think that would be confusing/weird we can have this update routine.

classes/Avif/RewriteRules/Nginx.php
wordpressfan: I already mentioned this also in the PR description :D 
to keep the structure consistent, I kept it.
But let's think about this together.

inc/functions/attachments.php
wordpressfan: Many thanks, going to remove it then, wasn't aware of this previous PR.

remyperona (approved):

Mai-Saad:
Rules applied correctly on apache, local and imagify.rocketlabs.

Note: on https://imagify.rocketlabsqa.ovh/ if we uploaded the zip file of any plugin, we will have error in media library (may needs further investigations when documenting changes needed at nginx server) => without playing with nginx rules at newer.rocketlabsqs upload plugin zip is fine
Screenshot from 2024-03-01 09-21-55

PR #828, opened 2024-02-28
Fixes #826 Only generate missing next-gen when enabling AVIF option

remyperona:
# Description

This PR changes the action happening when enabling the AVIF option. Instead of running the bulk optimize process, it runs the generate missing next-gen process, so already optimized or generated images should not be recreated for nothing.

Fixes #826

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue).

# Checklists

## Feature validation

- [x] I validated all the Acceptance Criteria. If possible, provide sreenshots or videos.
- [x] I triggered all changed lines of code at least once without new errors/warnings/notices.

## Code style
- [x] I wrote self-explanatory code about what it does.
- [x] I named variables and functions explicitely.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unecessary complexity.

PR #832, opened 2024-02-29
Fix/830 bulk optimisation

### LINKED ISSUE #830: Bulk optimization isnot working
- Made sure you’re on the latest version => feature/avif and fix/826-prevent-reoptimize

**Describe the bug**
When apply bulk optimization, message displayed that all images optimized although it isnot

**To Reproduce**
1. Fresh install/activation of feature/avif
2. upload some images (small, medium, large) => optimization is done
3. click restore original image for optimized image => next-gen removed and image(s) timestamp is updated
4. got to bulk optimization and click imagify all => message displayed that images are optimized
Screenshot from 2024-02-29 10-37-34

**Expected behavior**
Bulk optimization is applied as same as on trunk (image timestamp/size is updated + next-gen is added as per enabled option)

**Screenshots**

when click bulk optimization and display that images already optimized
Screenshot from 2024-02-29 10-51-05

after manually click on the optimize 
Screenshot from 2024-02-29 10-56-23

**Additional context**

Khadreal:
# Description

Fixes #830 

### User documentation

This allow users don't to bulk optimise images that aren't optimise.

### Technical documentation

*Explain how this code works. Diagram & drawings are welcomed.*

## Type of change

## New dependencies
N/A

## Risks
*List possible performance & security issues or risks, and explain how they have been mitigated.*

# Checklists

## Feature validation

- [ ] I validated all the Acceptance Criteria. If possible, provide sreenshots or videos.
- [ ] I triggered all changed lines of code at least once without new errors/warnings/notices.

## Documentation

- [ ] I prepared the user documentation for the feature/enhancement and shared it in the PR or the GitHub issue.
- [ ] I prepared the technical documentation if needed, and shared it in the PR or the GitHub issue.

## Code style

### DIFF
--- a/classes/Bulk/Bulk.php
+++ b/classes/Bulk/Bulk.php
@@ -21,13 +21,12 @@ class Bulk {
 		add_action( 'imagify_convert_webp_finished', [ $this, 'clear_webp_transients' ], 10, 2 );
 		add_action( 'wp_ajax_imagify_bulk_optimize', [ $this, 'bulk_optimize_callback' ] );
 		add_action( 'wp_ajax_imagify_missing_nextgen_generation', [ $this, 'missing_nextgen_callback' ] );
-		add_action( 'imagify_bulk_optimize', [ $this, 'bulk_optimize' ], 10, 2 );
 		add_action( 'wp_ajax_imagify_get_folder_type_data', [ $this, 'get_folder_type_data_callback' ] );
 		add_action( 'wp_ajax_imagify_bulk_info_seen', [ $this, 'bulk_info_seen_callback' ] );
 		add_action( 'wp_ajax_imagify_bulk_get_stats', [ $this, 'bulk_get_stats_callback' ] );
 		add_action( 'imagify_after_optimize', [ $this, 'check_optimization_status' ], 10, 2 );
 		add_action( 'imagify_deactivation', [ $this, 'delete_transients_data' ] );
-		add_action( 'update_option_imagify_settings', [ $this, 'maybe_bulk_optimize_callback' ], 10, 2 );
+		add_action( 'update_option_imagify_settings', [ $this, 'maybe_generate_missing_nextgen' ], 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Miraeld (approved):

Miraeld:
I've tried this scenario:

1 - Upload some images,
2 - Install Imagify
3 - Enable AVIF
4 - Start a bulk optimization 
5 - AVIF are generated. 

👍

classes/Bulk/Bulk.php
```diff
@@ -187,6 +186,9 @@ public function run_optimize( string $context, int $optimization_level ) {
 			$result = $this->get_bulk_instance( $context )->get_optimized_media_ids_without_format( $format );
 			$media_ids['ids'] = array_merge( $media_ids['ids'], $result['ids'] );
 		}
```
remyperona, line 188: Why are we looping on the formats and optimized media without formats? This is not the purpose of this method. I think only getting the unoptimized media is enough

remyperona (changes requested):

classes/Bulk/Bulk.php
Khadreal: That was added as part of the solution for avif that aren't automatically generated after enabling the option

@Miraeld I think you might want to expatiate further, since it was part of the #796

remyperona: I think it's not needed anymore after the change from latest PR #828

Miraeld: If I may, 
The problem of the method `get_optimized_media_ids_without_format($format)` is that the SQL query within this method only look for the media that have already been optimized by Imagify but without the format. 

So in the following scenario: 
1 - Upload some images,
2 - Install Imagify
3 - Enable AVIF
4 - Start a bulk optimization

It won't work, as the uploaded image have never been optimized by Imagify, so in the database we won't have any `_imagify_status` or `_imagify_data` within the `wp_postmeta` table. 

While the function `get_unoptimized_media_ids` doesn't care about these 2 metas values. And in the precedent scenario it will gets the IDs of these images and be able to convert them.

Mai-Saad:
Working fine now so if no further changes we can merge it.

classes/Bulk/Bulk.php
remyperona: With `get_unoptimized_media_ids()`, it will optimize the image and generate any next-gen image too at the same time. For me that's enough, and that is the behaviour that was in previous versions

MathieuLamiot: Thanks for the catch and details provided here @Tabrisrp.
Since the PR has been tested and the expected behavior is now OK, are you OK with opening a dedicated issue to try and remove the `get_optimized_media_ids_without_format` loop part?
This for loop might impact performances but that's only when doing bulk optimization so I think it's acceptable.
This would allow to move forward with the PR and 2.2 without having to re-test, and possible regressions for now ; and keep that optimization for the next Imagify minor (which should be soon anyway).

remyperona: follow up PR opened here #842

PR #836, opened 2024-03-04
Update 3rd-party compatibilities for multiple next-gen formats

remyperona:
# Description

Update some 3rd-party compatibilities to support the updates in 2.2 for next-gen formats

## Type of change

# Checklists

## Code style


Miraeld (approved):

Mai-Saad:
@Tabrisrp @piotrbak can you please share the scenarios related to compatibility with amp and regenerate thumbnail plugins :pray:

remyperona:
There is no functional code changes here, it's only filter and variables names change. But the cases are:
- for AMP, not display picture tag on AMP pages when the picture option is enabled
- for regenerate thumbnails, auto-optimize and generate next-gen for images that are regenerated

Mai-Saad:
@Tabrisrp Thanks for the feedback.
Exploratory test notes
**Amp**
Notes already on  2.1.3
- with Amp transitional mode, if we visited amp page while display webp using picture is enabled/disabled, only single image will be displayed (tested on page with image gallery of 3 images L, M, S  while display on , only Large image '1st image in media list' displayed) => **same with imagify deactivated**
- with standard setup of amp , visiting page?amp will redirect to page , however, images in source still in amp tag and not rewritten to image

**Woocommerce**
already on 2.1.3 (may need GH)
- For variable product, we rewrite product image to webp but not the one in the jquery.zoom.min.js. However, if we clear variations, the image in zoom request will be next-gen => shouldn't we unify the process here, either always rewrite image in zoom request or don't rewrite it (I think it's related to the fact that zoom won't be as efficient as it should when webp is used when clear variation selection, the image in zoom is rewritten to webp and in this case zoom won't be much)
Screenshot from 2024-03-15 13-28-55

- we arenot serving next-gen for Variant images (served from optimized not webP)
- if we uploaded new product image, processing in that page will keep loading although image is optimized

**Regenerate thumbnail**
Already on 2.1.3
- if we manually deleted next-gen files for certain thmbnail , then we regenerated thumbnail, the webP won't be created for those thumbnails (at thumbnail, this is checked `Skip regenerating existing correctly sized thumbnails (faster).`)

Mai-Saad (approved):
working as same as on trunk
[Uploading testrail-report-586.pdf…]()

PR #893, opened 2024-08-13
Adjust priority for template_redirect 

### LINKED ISSUE #891: Update callback priority on template_redirect for picture display
**Context**
Related to https://github.com/wp-media/wp-rocket/issues/6786

WP Rocket Lazyload exclusion is not working correctly when excluded pattern is part of a `picture` element created by the picture rewriting feature of Imagify.

**Expected beavhior**
Lazyload exclusion should work as expected

**Additional information**
So I believe I found what was happening with the next-gen issue. we have 2 callbacks on `template_redirect`:

- Imagify callback on priority -1000
- WP Rocket callback on priority 2

So we should expect imagify to be finished first, and WP Rocket after. BUT, both callbacks have an `ob_start()` call, and when multiple are called, they become nested until the buffering stops:

- Imagify opens buffer first
- then WP Rocket, nested inside

When output buffering ends (on shutdown action of WP), lower level content is passed to upper, so the order becomes WP Rocket optimizations first, then Imagify second, and HTML is output.
While callback priority is as expected, the output buffering works in the opposite order.

To verify, I tested by changing the priority of Imagify callback to 10, reverting both orders, and it works as expected, the picture element is applied first in the HTML, and then lazyload exclusion works as needed.

wordpressfan:
# Description

Fixes #891

### User documentation

As mentioned in the issue itself, here we fixed the priority of `template_redirect` hook's callback to be `10` instead of `-1000` to fix the buffer closing conflict with WP Rocket.

### Technical documentation

Just changed the priority.

## Type of change
*Delete options that are not relevant.*

## New dependencies

## Risks

As I mentioned here:
https://github.com/wp-media/imagify-plugin/issues/891#issuecomment-2283827408

We added this `-1000` priority in the mentioned PR to support rocket CDN so I tested this and it works so we may need to check another CDN plugin to see how it works with this branch.

# Checklists

## Feature validation

## Documentation

- [x] I prepared the user documentation for the feature/enhancement and shared it in the PR or the GitHub issue.
- [x] I prepared the technical documentation if needed, and shared it in the PR or the GitHub issue.

## Code style
- [ ] I wrote comments to explain why it does it.
- [ ] I named variables and functions explicitely.

## Observability
- [ ] I handled errors when needed.
- [ ] I wrote user-facing messages that are understandable and provide actionable feedbacks.
- [ ] I prepared ways to observe the implemented system (logs, data, etc.).

## Risks
- [x] I explicitely mentioned performance risks in the PR.
- [ ] I explicitely mentioned security risks in the PR.

### DIFF
--- a/classes/Picture/Display.php
+++ b/classes/Picture/Display.php
@@ -42,7 +42,7 @@ class Display implements SubscriberInterface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'template_redirect'            => [ 'start_content_process', -1000 ],
+			'template_redirect'            => 'start_content_process',
 			'imagify_process_webp_content' => 'process_content',
 		];
 	}


remyperona (approved):

Khadreal (approved):

Mai-Saad:
@wordpressfan thanks for the PR.
Here are the test results using imagify and WPR so far https://docs.google.com/spreadsheets/d/1YpkRpc_PkcszOezpkIooWC6hIWnkIWnXITvoHNjEjIo/edit?gid=0#gid=0
1- Are we going to handle the case in C6 here?
2- Is the o/p on C2 and C7 expected?
3- Do we need other GH about lcp loaded twice in NW and CDN not writing lcp totally? @piotrbak 

Note: for row 2 , the behavior is different if we used LL plugin instead of enable the option in WPR , with LL plugin the lcp won't be excluded from LL at all. any idea why LL plugin behaves differently than LL in WPR ? (test page https://new.rocketlabsqa.ovh/gallery/)

Mai-Saad:
@wordpressfan As per discussion with @piotrbak we need to handle the C6 related scenario
1- Enable display picture
2- Enable CDN, LL image
3- Visit page have no lcp yet => lcp is added to DB
4- Clear cache and revisit the page => lcp /atf shall be excluded from LL (currently it is not LL)

- CDN rewrite issues can be handled here https://github.com/wp-media/wp-rocket/issues/6652
- Loading LCP image twice can have a separate issue

remyperona:
Rocket lazyload plugin uses a priority of `2` on `template_redirect`, same as WP Rocket

PR #900, opened 2024-09-25
Closes #2883: Imagify Plugin - Replace the documentation link with the upgrade one for Free users

jeawhanlee:
# Description

Fixes https://github.com/wp-media/imagify/issues/2883
Users on starter plan will see link to upgrade from the plugin page.

## Type of change

## Detailed scenario

N/A

### Documentation

Replaces the documentation link with the upgrade link in the plugin action links if the plan_id = 1 

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style


Khadreal (approved):

inc/functions/common.php
```diff
@@ -238,90 +238,92 @@ function imagify_can_optimize_custom_folders() {
...
-function imagify_get_external_url( $target, $query_args = array() ) {
-	$site_url = IMAGIFY_SITE_DOMAIN . '/';
-	$app_url  = IMAGIFY_APP_DOMAIN . '/#/';
-
-	switch ( $target ) {
-		case 'plugin':
-			/* translators: Plugin URI of the plugin/theme */
-			$url = __( 'https://wordpress.org/plugins/imagify/', 'imagify' );
-			break;
+if ( ! function_exists( 'imagify_get_external_url' ) ) {
```
remyperona: what is the need of wrapping the function in this conditional?

jeawhanlee: Test is returning an error of the function already defined after mocking it.

remyperona: We should not be modifying the code to make the tests work, it would be good to look into how the code needs to be modified to be easily testable instead.

inc/classes/class-imagify-views.php
```diff
@@ -222,7 +222,17 @@ public function add_network_menus() {
 	 * @return array
 	 */
 	public function plugin_action_links( $actions ) {
```
remyperona: IMO, this is a good opportunity to move code to a new architecture more in line with our standards.

We could create a new class in `classes/Admin/`, `Admin.php` (for lack of a better name), implementing the Subscriber interface, and having `User` as a dependency instead of creating the instance in the method itself.

We can also add a service provider to instantiate it.

That will make things easier to test, and start a foundation for the future of the code related to the admin part

remyperona (changes requested):

inc/classes/class-imagify-views.php
jeawhanlee: You're right

remyperona (approved):

Mai-Saad:
@marija-nikolic Here are screenshots after the change (looks good to me)
- starter plan
Screenshot from 2024-10-01 14-40-52


- growth plan
growth

- infinite plan
infinite

Notes:
- upgrade link is working
- translations aren't handled here

marija-nikolic:
> @marija-nikolic Here are screenshots after the change (looks good to me)
> 
> * starter plan
>   Screenshot from 2024-10-01 14-40-52
> * growth plan
>   growth
> * infinite plan
>   infinite
> 
> Notes:
> 
> * upgrade link is working
> * translations aren't handled here

Thanks. Can we please update the color to #6f9c3b (to improve visibility)?

remyperona:
Updated screenshot

upgrade

marija-nikolic:
> Updated screenshot
> 
> upgrade

Thank you, it seems more visible now.

PR #899, opened 2024-09-25
Closes #2884 Upgrade plan link in admin bar

remyperona:
# Description

Fixes #2884

Add an upgrade plan link in the admin bar for free users with quota < 20%

## Type of change

- [x] Enhancement (non-breaking change which improves an existing functionality).

## Detailed scenario

- Free account with more than 20% quota remaining
- Display the imagify admin bar sub-menu
- Upgrade plan link is displayed
- Clicking on link opens the subscription popup

upgrade-plan

### Documentation

Display the button if free user and quota > 20%

# Mandatory Checklist

## Code validation

- [x] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.
- [x] I triggered all changed lines of code at least once without new errors/warnings/notices.

## Code style
- [x] I wrote a self-explanatory code about what it does.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unnecessary complexity.

PR #907, opened 2024-10-11
Cache user API call for 5 minutes

remyperona:
# Description

Cache the user API call for 5 minutes to avoid the request on each page load.

## Type of change

## Detailed scenario

- With current `develop`, using query monitor, you can see there is a request to `https://app.imagify.io/api/users/me/` on every page load
- Using this PR branch, the request happens once, and should not happen again for 5 minutes

### Documentation

Following changes made in #899, #900, #902, the `User` class is used on more pages than before, and this class makes an API call to get the user information, with a timeout of 10 seconds.

Caching the result of the API call allows to prevent the request on every page.

It was not possible to use the existing `imagify_get_cached_user()` function, because this function returns a cache that doesn't match the result from the API call, but a result from the cache of the `User` class itself.

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/Tests/Unit/inc/classes/ImagifyUser/getError.php
+++ b/Tests/Unit/inc/classes/ImagifyUser/getError.php
@@ -31,7 +31,7 @@ class Test_GetError extends TestCase {
 			'is_monthly'                   => true,
 		];
 
-		Functions\when( 'imagify_get_cached_user' )->justReturn( false );
+		Functions\when( 'get_transient' )->justReturn( false );
 		Functions\when( 'get_imagify_user' )->justReturn( $userData );
 
 		$this->assertFalse( ( new User() )->get_error() );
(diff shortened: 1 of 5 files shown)


Khadreal (approved):

wordpressfan:
@Tabrisrp 
Do u know what is the difference between this PR and the code found here:
https://github.com/wp-media/imagify-plugin/blob/5684433e2a7d126b312de47ed05d21022483374c/inc/functions/admin.php#L391-L424

Now we inroduce a transient called `imagify_user_cache` and there is another one called `imagify_user`

if it's OK to introduce this transient, should we update the data being saved to have those attributes?

https://github.com/wp-media/imagify-plugin/blob/5684433e2a7d126b312de47ed05d21022483374c/inc/functions/admin.php#L414-L415

remyperona:
I explained it in the technical description, they're not exactly the same data. One is caching the API call itself, the other is caching the User class.

Ideally we would implement things differently, but it would require too much work right now.

Mai-Saad:
On PR => refresh admin won't send API call to imagify till this transient is expired `imagify_check_api_version`
On trunk => API call is there with every refresh

PR #996, opened 2025-07-28
Exclude rules from CS temporarily

remyperona:
# Description

Exclude rules from CS temporarily

## Type of change

- [x] Chore

### What was tested

n/a

### How to test

n/a


nicomollet:
@remyperona Can we enable back the **rules from CS** (Imagify plugin)?

PR #1002, opened 2025-12-19
Closes #1001 Hide our plugins section for given partners

### LINKED ISSUE #1001: As a partner, I want to install imagify on websites with the "our plugins" tab disabled
**Context**

Some partners want to work with Imagify and pre-install it on their websites, using the partners_id mechanism, but without the "our plugins" page.

**Dependencies**

NA

**Expected behavior**

When the partners_id is set to value in a short-list (hardcoded, possibly with a filter), then the "our plugins" tab must not appear.

**Additional information**

The short-list can be given by the Marketing team (@EstherWPR, @Agathemed) on Slack

remyperona:
# Description

Fixes #1001 

## Type of change

### What was tested

- Tested hiding of the our plugins section when the option `imagify_partner_hide_our_plugins`is set

### How to test

There is 2 ways to test the flow:
- When creating a new user from the plugin, with the partner ID option `imagifyp_id` set
or
- If the user already exists, saving the API key, with the partner ID option `imagifyp_id` set

In both of those cases, if the partner ID is `extendify`, the Our Plugins section on the settings page should be hidden.

This is done through an option called `imagify_partner_hide_our_plugins` which is saved when deleting the partner ID option after one of the 2 above flow happened, if the partner corresponds to extendify (for now).

- Affected is the settings page, our plugins section

### Documentation

This PR implements functionality to hide the "Our Plugins" section in the settings page for specific partners. When users sign up through partner, the plugin family section will not be displayed to them.

Key changes:
- Adds a new function to save partner IDs that require hiding the plugins section
- Integrates this function into both API key setup and user creation flows
- Conditionally renders the "Our Plugins" section based on stored partner information

# Mandatory Checklist

## Code validation

## Code style

TBD


Mai-Saad:
@remyperona Thanks for the PR. Please find exploratory test notes below.
- https://github.com/wp-media/imagify-plugin/issues/1003 => same on develop
- If we use snippet to simulate partner 
`add_filter( 'pre_option_imagifyp_id', function() {
    return 'wp-rocket';
} );` ===> plugins section won't be displayed as expected :heavy_check_mark: 
- If we disabled the snippet above, then did fresh install of PR ==> the plugins section wouldn't be displayed, shouldn't it be displayed?

inc/functions/partners.php
```diff
@@ -36,3 +34,24 @@ function imagify_delete_partner() {
...
+ *
+ * @since 2.2.7
+ *
+ * @param string|bool $partner Partner ID.
+ *
+ * @return void
+ */
+function imagify_save_partner_hide_our_plugins( $partner ) {
+	$partners = [
+		18, // Extendify.
```
nicomollet: I wonder if we could use the slugs instead of IDs?

[ Metabase > Datator > Imagify Partners](https://data.wp-media.me/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJ0eXBlIjoicXVlcnkiLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjY3fX0sImRpc3BsYXkiOiJ0YWJsZSIsInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9fQ==)

Example of usage with string/slug:
https://github.com/wp-media/wp-rocket/blob/a9d6455ad5b9dcc3a5b707336cb76a1a5e681279/inc/Plugin.php#L198C37-L198C46

nicomollet (changes requested):
Feedback;:
- CI (coverage) not passing
- Partner ID or slug usage?
- We need `What was tested` and `How to test` please 🙏

inc/functions/partners.php
remyperona: You're right, in fact a string is expected, even though the option name is `imagifyp_id` so it's confusing, but it's old code.

nicomollet (approved):

Mai-Saad:
@remyperona Thanks for the update.
- If no snippet about partner, plugins section is displayed :heavy_check_mark: 
Screenshot from 2026-01-12 08-29-53

- If we have snippet about extendify then fresh install `add_filter( 'pre_option_imagifyp_id', function() {
    return 'extendify';
} );` => plugins section isnot displayed ====> we still need to remove separator before save buttons , as mentioned here https://group-onecom.slack.com/archives/C08ECU1CH6Z/p1767886913617889?thread_ts=1765980057.429929&cid=C08ECU1CH6Z , can you please check :pray: 
Screenshot from 2026-01-12 08-42-07

- If we disabled the snippet above, then did fresh install of PR, plugins section is displayed :heavy_check_mark:

hanna-meda:
@remyperona, @remyperona  confirming that with latest changes the separator before the save buttons is now removed. 
Screenshot 2026-01-12 at 17 57 58

PR #1007, opened 2026-01-12
Add Extendify compatibility class

remyperona:
# Description

Modify the compatibility with Extendify to use an option instead of the partner ID

## Type of change

### What was tested

Tested the display of the `Our Plugins` section, based on the existance of the option `extendify_site_id`

### How to test

You can simulate the option using:

```
add_filter( 'pre_option_extendify_site_id', function() {
	return 1;
});
```

Imagify settings page, our plugins section

### Documentation

This pull request refactors the Extendify compatibility mechanism from a partner ID-based system to an option-based approach. Instead of checking a custom `imagify_partner_hide_our_plugins` option, the plugin now checks for the existence of Extendify's own `extendify_site_id` option to determine whether to hide the "Our Plugins" section.

**Changes:**
- Removed the legacy partner-based hiding system including the `imagify_save_partner_hide_our_plugins` function and related calls
- Introduced a new `Extendify` subscriber class that hooks into the `imagify_hide_plugin_family` filter
- Replaced direct option checking with a filter-based approach using `wpm_apply_filters_typed`

# Mandatory Checklist

## Code validation

## Code style


Mai-Saad:
@remyperona Thanks for the PR. Here is exploratory test notes
- For the mentioned filter `add_filter( 'pre_option_extendify_site_id', function() {
	return 1;
});` any other values will work and hide the plugins section i.e -1, 0, 6, null ===> is this expected?
- If we disable the snippet, the plugins section will be displayed with no need to reinstall plugin, ok?
- When deactivate plugin, can see this console error, any idea how can we gaurd that?
Screenshot from 2026-01-13 11-10-08

- Waiting for confirmation if the current UI is ok or needs an update https://group-onecom.slack.com/archives/C08ECU1CH6Z/p1768288243780879?thread_ts=1765980057.429929&cid=C08ECU1CH6Z ==> It's not critical but it would definitely look better. @remyperona if nothing will be done here then we can merge this PR

nicomollet (approved):
Approving

For @Mai-Saad 
- Any value in `extendify_site_id` would hide the plugin family, it is expected.
- The error in screenshot `ERR_QUIC_PROTOCOL` doesn't seem related to Imagify: can you retry again?

Mai-Saad:
>The error in screenshot ERR_QUIC_PROTOCOL doesn't seem related to Imagify: can you retry again?

@nicomollet yes working fine now :pray:

PR #1019, opened 2026-02-05
Add PHPStan custom rule for `get_subcribed_events()` callbacks validation

remyperona:
# Description

Port the existing validations for callbacks of `add_filter()` and `add_action()` to the subscriber system and `get_subscribed_events()` array returnedé

## Type of change

- [x] Chore

## SubscriberCallbackRule

This rule validates `get_subscribed_events()` method implementations in classes that implement `SubscriberInterface`.

### Purpose

Ensures that:
- All array elements have `@filter` or `@action` annotations
- Filter callbacks return non-void values
- Action callbacks return void
- Parameter counts match `accepted_args` when specified

### Usage

Add single-line comment annotations to each array element in `get_subscribed_events()`:

```php
public static function get_subscribed_events(): array {
    return [
        // @action
        'wp_ajax_imagify_get_admin_bar_profile' => 'get_admin_bar_profile_callback',
        
        // @action
        'admin_bar_menu' => [ 'add_imagify_admin_bar_menu', IMAGIFY_INT_MAX ],
        
        // @filter
        'imagify_cdn_source_url' => 'get_cdn_source',
        
        // @filter
        'imagify_settings' => [ 'modify_settings', 10, 2 ],
    ];
}
```

### Supported Array Formats

1. **Simple string method name:**
   ```php
   // @action
   'hook_name' => 'method_name'
   ```

2. **Array with method and priority:**
   ```php
   // @filter
   'hook_name' => [ 'method_name', 20 ]
   ```

3. **Array with method, priority, and accepted_args:**
   ```php
   // @action
   'hook_name' => [ 'method_name', 10, 3 ]
   ```

4. **Nested arrays (multiple callbacks for same hook):**
   ```php
   // @action
   'hook_name' => [
       [ 'method_name_1', 10, 2 ],
       [ 'method_name_2', 20, 1 ]
   ]
   ```
   Note: Annotation on parent element applies to all nested callbacks.

5. **Variable assignment:**
   ```php
   public static function get_subscribed_events(): array {
       $events = [
           // @action
           'hook_name' => 'method_name',
       ];
       return $events;
   }
   ```

#### Filters (`// @filter`)
- **Must return a value** (cannot return void)
- Used when modifying data that flows through the hook
- Example:
  ```php
  // @filter
  'the_content' => 'modify_content',
  
  public function modify_content( string $content ): string {
      return $content . ' Modified!';
  }
  ```

#### Actions (`// @action`)
- **Must return void** (or explicit never)
- Used when performing side effects without returning data
- Example:
  ```php
  // @action
  'admin_init' => 'register_settings',
  
  public function register_settings(): void {
      register_setting( 'group', 'option' );
  }
  ```

### Error Identifiers

- `wpmedia.subscriber.missingAnnotation` - Missing @filter or @action annotation
- `wpmedia.subscriber.filter.missingReturn` - Filter callback doesn't return a value
- `wpmedia.subscriber.action.unexpectedReturn` - Action callback returns a value
- `wpmedia.subscriber.arguments.count` - Parameter count mismatch with accepted_args

### Skipped Validations

The rule automatically skips validation for:
- Empty array returns
- Delegated implementations (e.g., `return ParentClass::get_subscribed_events()`)
- Complex multi-statement array building
- Conditional array modifications

#### Valid Implementation

```php
class MySubscriber implements SubscriberInterface {
    public static function get_subscribed_events(): array {
        return [
            // @action
            'init' => 'initialize',
            
            // @filter
            'the_content' => [ 'filter_content', 10, 2 ],
        ];
    }
    
    public function initialize(): void {
        // Action - returns void
        register_post_type( 'my_type', [] );
    }
    
    public function filter_content( string $content, int $post_id ): string {
        // Filter - returns value
        return $content . ' Filtered';
    }
}
```

#### Invalid Implementation (Will Show Errors)

```php
class MySubscriber implements SubscriberInterface {
    public static function get_subscribed_events(): array {
        return [
            // Error: Missing annotation
            'init' => 'initialize',
            
            // @filter
            'the_content' => 'filter_content',
        ];
    }
    
    // Error: Action returns string instead of void
    public function initialize(): string {
        return 'initialized';
    }
    
    // Error: Filter returns void instead of value
    public function filter_content( string $content ): void {
        echo $content;
    }
}
```

```

### DIFF
--- a/Tests/phpstan/Rules/SubscriberCallbackRule.php
+++ b/Tests/phpstan/Rules/SubscriberCallbackRule.php
@@ -184,25 +184,28 @@ class SubscriberCallbackRule implements Rule
 	}
 
 	/**
-	 * Get hook type (@filter or @action) from array item PHPDoc comment
+	 * Get hook type (@filter or @action) from array item comment
 	 */
 	private function getHookType( ArrayItem $item ): ?string
 	{
-		$docComment = $item->getDocComment();
+		$comments = $item->getAttribute( 'comments' );
 
-		if ( $docComment === null ) {
+		if ( empty( $comments ) ) {
 			return null;
 		}
 
-		$text = $docComment->getText();
+		// Check all comments attached to this node
+		foreach ( $comments as $comment ) {
+			$text = $comment->getText();
 
-		// Match @filter or @action
-		if ( preg_match( '/@filter\b/i', $text ) ) {
-			return 'filter';
-		}
+			// Match // @filter or // @action
+			if ( preg_match( '/\/\/\s*@filter\b/i', $text ) ) {
+				return 'filter';
+			}
 
-		if ( preg_match( '/@action\b/i', $text ) ) {
-			return 'action';
+			if ( preg_match( '/\/\/\s*@action\b/i', $text ) ) {
+				return 'action';
+			}
 		}
 
 		return null;
--- a/Tests/phpstan/Rules/SubscriberCallbackRule.php
+++ b/Tests/phpstan/Rules/SubscriberCallbackRule.php
@@ -1,0 +1,468 @@
+<?php
+
+namespace Imagify\Tests\phpstan\Rules;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\ArrayItem;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Scalar\Int_;
+use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Return_;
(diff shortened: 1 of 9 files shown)


Miraeld (approved):

wordpressfan:
@remyperona do u think it'd be better to have this annotation on the method itself not into the subscriber array item?
Because sometimes we have one method for more than one hook.
Also for WPR we will reset the baseline.

remyperona:
One method can be on multiple hooks, but the same rules apply: a return for a method on a filter, no return for a method on an action.

So in the end, it is the hook that determines what should happen in the method, not the other way around.

wordpressfan:
> One method can be on multiple hooks, but the same rules apply: a return for a method on a filter, no return for a method on an action.
> 
> So in the end, it is the hook that determines what should happen in the method, not the other way around.

OK, thanks @remyperona 
So I believe we can merge this one.

PR #1085, opened 2026-06-12
feat(tools): add one-click Reset Internal State troubleshooting tool

### LINKED ISSUE #1012: Feature Request: One-Click Tool to Reset Imagify Internal State (Clear Transients and Stuck Jobs)
Context:

Optimization getting stuck is one of the most common technical issues in Imagify. While it can have multiple root causes (blocked connections, low memory, server limitations, etc.), in a large number of cases everything on the server side is actually fine.

Despite that, the UI continues to show issues such as:
	•	Endless optimization loading
	•	Optimization stuck at a certain percentage
	•	False “out of quota” error messages

In most of these cases, when we log in to the user’s site, the issue is resolved simply by deactivating and reactivating the plugin, or by reinstalling the plugin. After that, optimization starts working normally again.

This clearly indicates that the problem is often related to stale internal state, transients, or cached data rather than a real server or account issue.

Although we often have to log in to the user’s site to fix the issue, asking users as a first step to delete transients manually or to reinstall the plugin is not a great experience.  These steps feel risky and uncomfortable for users and are not something we should rely on as a primary troubleshooting action.

Instead, we could provide a dedicated button, either directly inside the Imagify plugin or inside Imagify Tools (even if it requires a separate installation), that performs the same internal cleanup that happens when the plugin is deactivated or reinstalled.

This would be similar to tools used in page builders, such as “Regenerate CSS” or cache purge actions, where users can safely reset internal data without breaking anything.

This is one of the issues I already raised with Nicolas during the Retreat.

Based on the code behavior, this is what currently happens when the plugin is deactivated or uninstalled.

Image

Image

Ideally, this button should not delete user settings and should not require the user to go through the setup again. It should only reset internal data, transients, and job-related state needed for optimization to work correctly.

This button could serve as a simple and safe first troubleshooting step for users. It would help resolve many “optimization stuck” cases on its own and would significantly reduce the number of support tickets.

I am attaching a few examples (out of many) where reinstalling or reactivating the plugin resolved the issue.
https://secure.helpscout.net/conversation/3104976053/591955
https://secure.helpscout.net/conversation/3196531155/614656
https://secure.helpscout.net/conversation/3103713746/591646
https://secure.helpscout.net/conversation/3103140053/591550

Additional Information:

In a few cases where everything looks fine at the server level and even reinstalling the plugin does not help, there is one more workaround that has helped users resolve the issue.

As a last step, we sometimes ask users to delete the imagify_optimize_media_batch rows from the wp_options table in the database. After doing this, optimization starts working again.

I am not fully sure whether this behavior is already tied to any existing uninstall or deactivation logic, or if this is a completely separate case that only affects the database state. The recurrence of this issue is not as regular as the other “optimization stuck” cases, but it does happen occasionally.

If this is related and technically safe, it might be worth considering whether this cleanup can also be handled by the same one-click troubleshooting button. That said, this would need careful validation to avoid any unintended side effects.

Attaching a few recent tickets where this workaround helped resolve the issue.
https://secure.helpscout.net/conversation/3131592480/598132
https://secure.helpscout.net/conversation/3112646772/593564
https://secure.helpscout.net/conversation/3192961069/613492
https://secure.helpscout.net/conversation/3135232978/599003

Miraeld:
> [!NOTE]
> 🤖 This PR was generated by the AI delivery pipeline (orchestrator · Claude Sonnet 4.6). All code was reviewed by the lead-reviewer and QA agents before this PR was marked ready.

# Description

Closes #1012

Adds a **Troubleshooting** section to the Imagify Settings page with a "Reset Internal State" button. Clicking the button POSTs to a new AJAX endpoint that clears stale optimization transients, process locks, and pending ActionScheduler jobs — the same cleanup that occurs on plugin deactivation, without touching user settings or media optimization records.

This resolves the most common "optimization stuck" support case with a single click, without requiring deactivate/reactivate.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

**Manual scenario 1 — Troubleshooting section visible with valid API key:**
- Navigated to Settings page with a valid API key → Troubleshooting section with "Reset Internal State" button is visible below the Our Plugins section and above the submit button.

**Manual scenario 2 — Section absent without API key:**
- Navigated to Settings page without an API key configured → Troubleshooting section does not render (intentionally inside `is_api_key_valid()` block).

**Manual scenario 3 — JS handler wires up correctly:**
- Verified `#imagify-reset-internal-state` button has a non-empty `data-nonce` attribute via browser DevTools.
- Verified `imagifyOptions.resetInternalState` is available in the page JS via browser console.

**Manual scenario 4 — Nonce key is `_wpnonce`:**
- Inspected POST payload in Network tab when clicking the button → `_wpnonce` key present (not bare `nonce`).

**Manual scenario 5 — Build passes:**
- `npm run build` succeeds with no errors. `assets/js/options.min.js` rebuilt from source.

**Manual scenario 6 — PHPCS passes:**
- `composer phpcs -- views/part-settings-tools.php inc/functions/i18n.php views/page-settings.php` returns 0 errors.

### How to test

1. Activate the plugin and ensure a valid API key is configured.
2. Go to **Settings → Imagify**.
3. Scroll to the bottom — a **Troubleshooting** section should be visible above the footer buttons.
4. Click **Reset Internal State**.
5. A SweetAlert confirmation dialog should appear.
6. Confirm → the button disables briefly, then a success message appears next to the button.
7. Verify in the browser Network tab that the POST goes to `admin-ajax.php` with `action=imagify_reset_internal_state` and `_wpnonce=<token>`.
8. Verify transients are cleared: with `WP_DEBUG` on, running a forced bulk optimization then resetting should unblock it.

- **Settings page** — new section added; existing settings sections and save button are unaffected.
- **AJAX handler** — new `imagify_reset_internal_state` endpoint (capability-checked, nonce-verified).
- **options.js** — new click handler added; existing handlers (API key check, backup alert, files tree) are unchanged.
- **uninstall.php** — already updated by backend agent to use `InternalStateList`; no frontend impact.

### Documentation

The full technical documentation is in `docs/api/tools.md`:
- AJAX action `imagify_reset_internal_state` with nonce action `imagify_reset_internal_state`.
- Nonce delivered via `data-nonce` attribute on the button (same pattern as `backup-dir-is-writable`).
- Nonce posted as `_wpnonce` in the request body — required by `imagify_check_nonce()` / `check_ajax_referer()`.
- JS localization in `imagifyOptions.resetInternalState` (action, confirm text, success/error labels).

### New dependencies

None. No new PHP or JS dependencies. SweetAlert and jQuery already loaded with the `options` script handle.

### Risks

- **Security:** Nonce-verified + capability-checked (`imagify_get_context('wp')->current_user_can('manage')`). No unauthenticated access possible.
- **Data loss:** Explicitly excludes user settings, API key, and all media optimization records. Only transients and scheduled-action rows are affected.
- **Double-click:** Button is disabled in-flight; server-side is idempotent (deleting absent transients is a no-op).
- **Multisite:** Sitemeta DELETE is guarded by `is_multisite()`; reset is scoped to current site only. Documented in PR and `docs/api/tools.md`.

## Acceptance Criteria

1. A "Troubleshooting" section with a "Reset Internal State" button appears on the Settings page for users with a valid API key
2. Clicking the button triggers an AJAX call that clears optimization-state transients, cancels pending ActionScheduler jobs, and removes process locks
3. User settings, API key, and media optimization records are NOT affected by the reset
4. Only logged-in admins with `manage_options` capability can trigger the reset
5. Nonce-protected AJAX endpoint validates properly
6. Button shows confirmation dialog before resetting, and displays success/error feedback
7. uninstall.php uses the shared InternalStateList static methods (single source of truth)
8. No inline CSS or inline JS in views

# Mandatory Checklist

## Code validation

## Code style

- [ ] Output messages (errors, notices, logs) are explicit enough for users to understand the issue and are actionnable.

- **Output messages** — The AJAX endpoint returns `wp_send_json_success()` on success and `wp_send_json_error()` on failure. User-facing feedback is in the JS layer (success/error strings in `imagifyOptions.resetInternalState`). No raw server-side log messages are exposed to end users; the feedback strings are translatable and human-readable.

# Additional Checks
- [ ] In the case of complex code, I wrote comments to explain it.


classes/Tools/ResetInternalState.php
```diff
@@ -0,0 +1,62 @@
...
+
+		// 2. Delete process-lock and legacy RPC transients via LIKE patterns in wp_options.
+		foreach ( InternalStateList::get_locked_transient_patterns() as $pattern ) {
+			$wpdb->query( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+				$wpdb->prepare(
+					"DELETE FROM {$wpdb->options} WHERE option_name LIKE %s", // phpcs:ignore WordPress.DB.PreparedSQLPlaceholders.LikeWildcardsInQuery
+					$pattern
+				)
+			);
+		}
```
Miraeld, line 45: [HIGH] The patterns returned by `InternalStateList::get_locked_transient_patterns()` are pre-escaped SQL LIKE sequences (e.g. `\_transient\_%imagify-auto-optimize-%`). When passed as `%s` to `$wpdb->prepare()`, wpdb quotes and escapes the string value but does NOT add a SQL LIKE escape clause. MySQL will interpret `\_` as an escaped literal underscore only under default MySQL config; under `NO_BACKSLASH_ESCAPES` mode the backslash escape sequences break. The original `uninstall.php` used raw string interpolation where those backslashes went directly into the SQL literal. The safe, idiomatic WordPress pattern is: build the pattern with `$wpdb->esc_like()` on the literal prefix/suffix, then concatenate `%` wildcards — see `inc/admin/upgrader.php:263`. The unit test mocks `prepare` so it does not catch this.\n\n**Fix:** Change `get_locked_transient_patterns()` to return plain prefix strings and in `reset()` build the pattern as `$wpdb->esc_like( $prefix ) . '%'` before passing to `prepare()`.

classes/Tools/Subscriber.php
```diff
@@ -0,0 +1,63 @@
...
+	 *
+	 * Note: the settings-section template is rendered directly via print_template()
+	 * in page-settings.php — no imagify_settings_tools hook is registered here.
+	 *
+	 * @return array
+	 */
+	public static function get_subscribed_events(): array {
+		return [
+			// @action
+			'wp_ajax_imagify_reset_internal_state' => 'handle_reset',
```
Miraeld: [MEDIUM] Subscriber callback method name deviates from spec.

The spec (Implementation Plan step 3) requires: `'wp_ajax_imagify_reset_internal_state' => 'reset_internal_state'` and a corresponding `public function reset_internal_state(): void` method. The implementation maps to `'handle_reset'` with method `handle_reset()`.

**Fix:** Rename the method from `handle_reset()` to `reset_internal_state()` and update the `get_subscribed_events()` return value to match. Update the Subscriber unit test (line 26) to assert `'reset_internal_state'` instead of `'handle_reset'`. This brings the code in line with the spec, the naming convention used by other AJAX callbacks in this codebase (e.g. `imagify_check_backup_dir_is_writable_callback`), and the hook name it serves.

Miraeld:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · claude-sonnet-4-6).

**Review: ❌ CHANGES REQUESTED**

**Blockers:**
- [MEDIUM] `classes/Tools/Subscriber.php:40` — AJAX callback method is named `handle_reset` / mapped as `'handle_reset'`, but the spec (Implementation Plan step 3) requires the method and mapping key to be `reset_internal_state`. Fix: rename `handle_reset()` to `reset_internal_state()`, update `get_subscribed_events()` return value to `'wp_ajax_imagify_reset_internal_state' => 'reset_internal_state'`, and update the Subscriber unit test assertion on line 26 accordingly. Naming must match the spec so the implementation is auditable and consistent with the hook name convention used in the rest of the codebase.

**Nice-to-haves:**
- `docs/api/tools.md` — file is not listed in the spec's Affected Files, but its content is accurate and useful; no action required unless the team prefers to track documentation changes in spec.

Miraeld (commented):
Fixed: renamed `handle_reset()` → `reset_internal_state()` in `classes/Tools/Subscriber.php` and updated `get_subscribed_events()` mapping + unit test assertion to match. Commit: 497c2fcf. All 12 unit tests pass locally.

classes/Tools/Subscriber.php
Miraeld: Fixed in 497c2fc — renamed `handle_reset()` → `reset_internal_state()` in `classes/Tools/Subscriber.php` and updated the subscriber map and unit test to match.

classes/Tools/ResetInternalState.php
Miraeld: Fixed in ccfaac3a — `get_locked_transient_patterns()` now returns plain (unescaped) strings; `ResetInternalState::reset()` assembles the final LIKE pattern via `esc_like()` on each segment separated by `%` wildcards. `phpunit.xml.dist` coverage include now covers `classes/` and a new `Test_ResetInternalState` test covers both `Subscriber::reset_internal_state()` paths.

Honemo:
@Miraeld 
image

The text is very compact in the popup. It might be worth spacing it out a bit by adding something like:
```
line-height: 1;
```
to the CSS.

Miraeld:
Screenshot 2026-06-15 at 16 24 02

Honemo (approved):

PR #1177, opened 2026-07-07
add filter to the return of mimes in imagify_get_mime_types.

mauroTabsSpaces:
# Description

## Type of change
-Enhancement (non-breaking change which improves an existing functionality).

## Detailed scenario
This allows developers to hook in on the MIME types of Imagify. With this hook you can add and remove MIME types so the will not be processed by upload or bulk optimization.

### What was tested
We tested the hook in in a php 8.4 wordpress setup. We hooked in to the hook and added and removed MIME like pdf and gif. After removing them they where no longer proccesed by the optimization and the bulk optimization.

### How to test
Create a hook in your current on imagify_get_mime_types and remove a MIME type like pdf. Then upload a pdf to the media library.

As far as we can see none real thing will changes except the possibility of removing a mime type and give people more freedom to to say which mime types the want to optimization.

### Documentation
With this change, we enable developers to add and remove MIME types, allowing them to specify more precisely what they want to optimize.

### New dependencies
None

### Risks
- People adding MIME types that are not working in the optimzions self.

### DIFF
--- a/inc/functions/attachments.php
+++ b/inc/functions/attachments.php
@@ -26,7 +26,7 @@ function imagify_get_mime_types( $type = null ) {
 		$mimes['pdf'] = 'application/pdf';
 	}
 
-	return $mimes;
+	return apply_filters( 'imagify_get_mime_types', $mimes );
 }
 
 /**


inc/functions/attachments.php
```diff
@@ -26,7 +26,7 @@ function imagify_get_mime_types( $type = null ) {
 		$mimes['pdf'] = 'application/pdf';
 	}
 
-	return $mimes;
+	return apply_filters( 'imagify_get_mime_types', $mimes );
```
Miraeld, line 29: Hello, 
In Imagify we are using a specific function instead of `apply_filters` 

Please take a look at https://github.com/wp-media/apply-filters-typed as this is the standard we use in all our projects 


```suggestion
	return wpm_apply_filters_typed( 'array', 'imagify_get_mime_types', $mimes );
```

Miraeld:
Hi @mauroTabsSpaces, thanks a lot for the contribution! 🙏

We've adopted your change in #1184 with two adjustments:
- Converted `apply_filters()` to `wpm_apply_filters_typed( 'array', ... )` per our project standard ([apply-filters-typed](https://github.com/wp-media/apply-filters-typed)), as mentioned in the review.
- Added the filter docblock.

You're credited as commit co-author, so the contribution stays yours. The reason for moving it to an internal branch is that our integration test suite requires repository secrets (an Imagify API key) that GitHub doesn't expose to fork PRs — so CI could never go green here regardless of the code.

Closing this in favor of #1184. Thanks again!

PR #1189, opened 2026-07-20
Fix wrong orientation of next-gen images generated from EXIF-rotated backups (#859)

### LINKED ISSUE #859: WebP generation of rotated JPG files results in wrong orientation
**Describe the bug**
WebP generation of rotated JPG files results in wrong orientation.

**To Reproduce**
1. Upload a JPG file with EXIF orientation data. The file below has orientation : 6 (Rotate 90 CW).

bird-box-portrait

**Expected behavior**
The WebP file would be rotated as per the JPG equivalent.

**Screenshots**
https://redbot.uk/imagify-bug-demo.html

- The 'full' sized WebP has the wrong orientation (see: bird-box-portrait-rotated.jpg.webp). 
- Resized WebP files have the correct orientation (see: bird-box-portrait-320x426.jpg.webp).

**Additional context**
On upload Wordpress rotates the JPG image via the maybe_exif_rotate() function, appending '-rotated' to the filename, before updating metadata orientation to 1.

- Raw fullsize path: .../wp-content/uploads/bird-box-portrait-rotated.jpg (orientation : 1).
- Backup path: .../wp-content/uploads/backup/bird-box-portrait.jpg (orientation : 6).

Imagify generates the 'full' size WebP file from the backup path, but does not replicate the rotation of the file performed on the raw fullsize path by Wordpress.

**Bug Fix**
The bug can be patched by adding the following code to the start of the optimize() function within file: /imagify/classes/Optimization/File.php

	public function optimize( $args = [] ) {
		$exif_meta = wp_read_image_metadata( $this->path );
		if ( ! empty( $args['convert'] ) && 'webp' === $args['convert'] && ! empty( $exif_meta['orientation'] ) && 1 !== (int) $exif_meta['orientation'] ) {
			$editor = wp_get_image_editor( $this->path );
			if ( ! is_wp_error( $editor ) ) {
				$rotated = $editor->maybe_exif_rotate();
				if ( true === $rotated ) {
					$editor->save( $this->path );
					// update metadata - orientation : 1 (not needed as removed for webp)
				}
			}
		}
		...
	}

Miraeld:
# Description

Fixes #859

When WordPress uploads a JPEG with a non-default EXIF orientation, it auto-rotates the pixels for the working (attached) file and resets that file's orientation to `1`. Imagify's backup, however, keeps a copy of the *original*, un-rotated JPEG (still carrying its original `Orientation` EXIF tag), because that's what's needed to losslessly restore the media later.

This becomes a problem when Imagify generates a Next-Gen (WebP/AVIF) version of the **full size** for a media that was already optimized (non-Next-Gen) beforehand: instead of re-deriving the Next-Gen file from the already-optimized full-size file (which would be re-compressing an already-lossy JPEG), Imagify creates a temporary working copy **from the backup** and converts that. Since the backup is the pre-rotation original, the resulting WebP/AVIF file ends up with the wrong orientation, even though the full-size JPEG on disk (and all its resized thumbnails, which are derived from it) display correctly.

This PR corrects the EXIF orientation of that temporary, disposable working copy right after it's created from the backup — before it's used to generate the Next-Gen version (or a thumbnail) — so the output always matches the orientation of the real, on-disk full-size file. **The backup file itself is never modified.**

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated only (see "Unticked items justification" below for the manual gap):

- New unit tests in `Tests/Unit/classes/Optimization/File/MaybeCorrectExifOrientationTest.php` cover `Imagify\Optimization\File::maybe_correct_exif_orientation()`:
  - no-op when EXIF can't be read, when the file isn't a JPEG, or when the orientation is already `1` (this last case is also the guard that prevents double-rotating the same working copy on repeated runs).
  - correct rotate/flip calls for orientation `5` and `6` (representative of the rotate-only and rotate+flip code paths), followed by `save()` over the same (temporary) path.
  - `WP_Error` from `wp_get_image_editor()` or from `$editor->save()` is surfaced without side effects.
- Full unit suite (`composer test-unit`): 318 tests, 0 failures (17 pre-existing risky-test notices, unrelated to this change).
- `composer run-stan`: 0 errors.
- `vendor/bin/phpcs --standard=phpcs.xml` on changed files: 0 errors/warnings.

### How to test

1. Upload a JPEG with EXIF orientation `6` (rotated 90° on capture) to the Media Library, with WebP and/or AVIF "Display" format enabled in Imagify settings, and with backup enabled.
2. Optimize the media normally first (non-Next-Gen), then trigger Next-Gen generation for it afterwards (e.g. via "Generate Next-Gen versions" bulk action, or by enabling the Next-Gen format after the fact and re-optimizing/completing the process) so that the code path that derives the Next-Gen full size from the backup is exercised.
3. Verify the resulting full-size `.webp`/`.avif` file has the same (correct) orientation as the full-size `.jpg` on disk — not the orientation of the pre-upload original.
4. Compute a checksum (e.g. `md5sum`) of the backup file (`wp-content/uploads/backup/.../image.jpg`) before and after the optimization/Next-Gen generation runs, and confirm it is unchanged — the backup must never be modified by this fix.

- Next-Gen (WebP/AVIF) generation for the **full** media size, specifically when it is (re)generated from the backup (i.e. the non-Next-Gen full size was already successfully optimized beforehand). This is the same code path used to create resized thumbnails from the backup, so thumbnail generation from the backup benefits from the same correction.
- Backup/restore integrity: the backup file must remain byte-identical.
- Re-optimize flows: repeated re-optimization from the backup must keep producing correctly-rotated output and must never double-rotate (guarded by only acting when EXIF orientation ≠ 1).

### Documentation

Root cause: `Imagify\Optimization\Process\AbstractProcess::create_temporary_copy()` copies the backup file to a disposable temporary path (`...@imagify-tmp.<ext>`) whenever a size's non-Next-Gen version was already successfully optimized and a Next-Gen (or thumbnail) version needs to be derived from an unmodified source. That temporary copy — not the backup itself — is what gets resized/converted by `Imagify\Optimization\File`.

The fix adds `Imagify\Optimization\File::maybe_correct_exif_orientation()`, which:
- only acts on JPEGs, and only when `exif_read_data()` is available (same guard already used by `File::resize()`),
- reads the `Orientation` EXIF tag of `$this->path`; does nothing if it's `1` (already upright, or not an EXIF-rotatable value — this is also what prevents double-rotation on repeated calls),
- otherwise gets a `WP_Image_Editor` instance for `$this->path`, applies the rotate/flip sequence (extracted into a new shared `rotate_editor_by_exif_orientation()` helper, reused by the existing `resize()` logic to avoid duplicating the orientation switch/case), and saves the result back over the same path.

`AbstractProcess::create_temporary_copy()` now calls `$tmp_file->maybe_correct_exif_orientation()` right after copying the backup into the temporary path, before that temporary file is used to generate the full-size Next-Gen version or a thumbnail. Because this only ever operates on the disposable `@imagify-tmp` copy (never on `$backup_path`), the backup file is guaranteed to stay untouched.

### New dependencies

None.

### Risks

- **Backup mutation risk**: mitigated by construction — `maybe_correct_exif_orientation()` is only ever called on the `File` instance wrapping the temporary copy (`$tmp_file`, pointing at `$tmp_path`), never on a `File` instance wrapping the backup path itself.
- **Double-rotation risk** (e.g. on repeated re-optimize runs): mitigated by the orientation check — once a temporary copy has been rotated and saved, its EXIF orientation is no longer `6` (or whichever original value), so a subsequent call is a no-op. In addition, `create_temporary_copy()` skips regenerating the temporary copy altogether if it already exists on disk.
- **Non-JPEG / no-EXIF-support environments**: the method safely no-ops (matches the existing behavior/guards already used by `File::resize()`), so no new failure mode is introduced there.
- **Editor/save failures**: surfaced as `WP_Error` but not currently treated as fatal by the caller (the temporary copy is used as-is if rotation fails) — this keeps behavior consistent with "no worse than before" rather than blocking the whole optimization pipeline on an EXIF-correction failure.

# Mandatory Checklist

## Code validation

- [ ] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.

## Code style

- Manual, real-upload validation with an actual EXIF-6 JPEG and a live WP install (checking the produced WebP/AVIF orientation visually, and diffing the backup file's checksum) was **not performed in this session** (no WordPress runtime available in this environment). Automated coverage is limited to unit tests exercising `File::maybe_correct_exif_orientation()` in isolation (mocked filesystem/editor); the integration suite (which would exercise the real `AbstractProcess::create_temporary_copy()` flow end-to-end) is known-broken locally (`WP_TESTS_DIR` missing) and was not run. Please perform the manual "How to test" steps above before/while reviewing.

# Additional Checks

### DIFF
--- a/Tests/Unit/classes/Optimization/File/MaybeCorrectExifOrientationTest.php
+++ b/Tests/Unit/classes/Optimization/File/MaybeCorrectExifOrientationTest.php
@@ -1,0 +1,192 @@
+<?php
+declare(strict_types=1);
+
+namespace Imagify\Tests\Unit\classes\Optimization\File;
+
+use Imagify\Optimization\File;
+use Imagify\Tests\Unit\TestCase;
+use Mockery;
+use WP_Error;
+
+/**
+ * Tests for \Imagify\Optimization\File::maybe_correct_exif_orientation().
+ *
+ * @covers \Imagify\Optimization\File::maybe_correct_exif_orientation
+ * @covers \Imagify\Optimization\File::rotate_editor_by_exif_orientation
+ * @group  Optimization
+ */
+class MaybeCorrectExifOrientationTest extends TestCase {
+
+	/**
+	 * Path used for the fake file under test.
+	 *
+	 * @var string
+	 */
+	private $path = '/tmp/fake-image@imagify-tmp.jpg';
+
+	/**
+	 * Builds a File instance with a mocked filesystem (and optionally editor) injected, bypassing
+	 * WordPress/filesystem globals.
+	 *
+	 * @param \Mockery\MockInterface $filesystem Mock for the Imagify_Filesystem dependency.
+	 * @param mixed                  $editor     Optional. Value to inject as the cached editor.
+	 * @param object|null            $file_type  Optional. Value to inject as the cached file type.
+	 *
+	 * @return File
+	 */
+	private function make_file( $filesystem, $editor = null, $file_type = null ): File {
+		// Bypass the real constructor: it instantiates Imagify_Filesystem::get_instance(), which
+		// requires a full WordPress environment that isn't available in unit tests.
+		$file = ( new \ReflectionClass( File::class ) )->newInstanceWithoutConstructor();
+
+		$this->setPropertyValue( 'path', $file, $this->path );
+		$this->setPropertyValue( 'filesystem', $file, $filesystem );
+
+		if ( null !== $file_type ) {
+			$this->setPropertyValue( 'file_type', $file, $file_type );
+		} else {
+			$this->setPropertyValue(
+				'file_type',
+				$file,
+				(object) [
+					'ext'  => 'jpg',
+					'type' => 'image/jpeg',
+				]
+			);
+		}
+
+		if ( null !== $editor ) {
+			$this->setPropertyValue( 'editor', $file, $editor );
+		}
+
+		return $file;
+	}
+
+	/**
+	 * When EXIF data can't be read at all, nothing should happen.
+	 */
+	public function testReturnsFalseWhenExifIsNotAvailable(): void {
+		$filesystem = Mockery::mock();
+		$filesystem->shouldReceive( 'can_get_exif' )->andReturn( false );
+		$filesystem->shouldNotReceive( 'get_image_exif' );
+
+		$file = $this->make_file( $filesystem );
+
+		$this->assertFalse( $file->maybe_correct_exif_orientation() );
+	}
+
+	/**
+	 * When the file isn't a JPEG, EXIF orientation is irrelevant.
+	 */
+	public function testReturnsFalseWhenNotJpeg(): void {
+		$filesystem = Mockery::mock();
+		$filesystem->shouldReceive( 'can_get_exif' )->andReturn( true );
+		$filesystem->shouldNotReceive( 'get_image_exif' );
+
+		$file = $this->make_file(
+			$filesystem,
+			null,
+			(object) [
+				'ext'  => 'png',
+				'type' => 'image/png',
+			]
+		);
+
+		$this->assertFalse( $file->maybe_correct_exif_orientation() );
+	}
+
+	/**
+	 * When the orientation is already 1 (or absent), the file must not be touched: this also
+	 * guards against rotating the same working copy twice.
+	 */
+	public function testReturnsFalseWhenOrientationIsAlreadyOne(): void {
+		$filesystem = Mockery::mock();
+		$filesystem->shouldReceive( 'can_get_exif' )->andReturn( true );
+		$filesystem->shouldReceive( 'get_image_exif' )->with( $this->path )->andReturn( [ 'Orientation' => 1 ] );
+
+		$editor = Mockery::mock();
+		$editor->shouldNotReceive( 'rotate' );
+		$editor->shouldNotReceive( 'flip' );
+		$editor->shouldNotReceive( 'save' );
+
+		$file = $this->make_file( $filesystem, $editor );
+
+		$this->assertFalse( $file->maybe_correct_exif_orientation() );
+	}
+
+	/**
+	 * Orientation 6 (the common "rotated 90° CW on capture" case) must rotate the editor by 270°
+	 * and save the result back over the same (temporary) path.
+	 */
+	public function testRotatesAndSavesForOrientationSix(): void {
+		$filesystem = Mockery::mock();
+		$filesystem->shouldReceive( 'can_get_exif' )->andReturn( true );
+		$filesystem->shouldReceive( 'get_image_exif' )->with( $this->path )->andReturn( [ 'Orientation' => 6 ] );
+
+		$editor = Mockery::mock();
+		$editor->shouldReceive( 'rotate' )->once()->with( 270 );
+		$editor->shouldNotReceive( 'flip' );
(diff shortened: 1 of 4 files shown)


remyperona:
WHile I think the fix works, I'm wondering if it's the best approach? Shouldn't Imagify backup the rotated image? It seems to be more likely the expected from an user perspective. If you restore to original a rotated image, is it reverting to the unrotated state currently?

Miraeld:
@remyperona 
Good question. I checked the restore path: it copies the backup (unrotated pixels plus the EXIF orientation tag) back over the full size, then regenerates thumbnails. WP rotates thumbnails correctly on load and browsers honor the tag on the full size, so it displays upright. So restore reverts to the true uploaded original, which is the intended lossless contract.

Backing up the rotated version would break that. WP's rotation is a lossy re-encode, so the backup would no longer be the pristine original. On top of that, every backup already on disk is pre-rotation, so we'd end up with mixed semantics and no migration path.

This PR keeps the backup byte identical and only fixes the disposable temp copy that next-gen is derived from. That's the narrower and safer fix.

Tests/Unit/classes/Optimization/File/MaybeCorrectExifOrientationTest.php
remyperona, line 1: This should be using dataProvider and fixtures file as we do usually

remyperona (changes requested):

remyperona (approved):

Miraeld:
## Manual QA — confirmed working ✅

Tested against this PR's exact commit (`92ad6325`) on a local WP install.

**Test setup**
- Generated a synthetic JPEG with a distinctive quadrant pattern (red / green / blue / yellow) and EXIF `Orientation: 6` baked in, so any wrong rotation is immediately obvious from raw pixel data (not just visual/EXIF-corrected previews).
- Enabled backup + WebP Next-Gen display in Imagify settings.

**Steps**
1. Uploaded the test JPEG to the Media Library. WordPress core auto-rotated it on upload, producing `exif-orientation-6-test-rotated.jpg` (corrected pixels, `Orientation: 1`) as the working full size, while leaving the raw original (`Orientation: 6`) on disk — this is standard WP 5.3+ behavior, not related to this fix.
2. Ran a normal (non-Next-Gen) optimization on the image first.
3. Then triggered Next-Gen (WebP) generation afterward, so Imagify derives the Next-Gen full size from the **backup** rather than the already-optimized working file — the exact code path this PR touches.
4. Compared raw pixel data (via GD, bypassing any EXIF auto-correction in viewers) between the corrected full-size JPEG and the generated WebP:

   | | top-left | top-right | bottom-left | bottom-right |
   |---|---|---|---|---|
   | `exif-orientation-6-test-rotated.jpg` | red | green | blue | yellow |
   | `exif-orientation-6-test-rotated.jpg.webp` | red | green | blue | yellow |

   Orientation matches exactly — the WebP is correctly rotated.
5. Verified backup integrity:
   ```
   md5sum wp-content/uploads/backup/2026/07/exif-orientation-6-test.jpg
   ```
   Checksum was identical before and after the optimize + Next-Gen generation run, and matches the raw original upload (`Orientation: 6` still intact) — the backup was never modified, as claimed in the PR description.

**Result:** Next-Gen files generated from the backup now come out correctly oriented, and the backup itself stays byte-for-byte untouched. Works as expected. 👍

PR #1190, opened 2026-07-20
Fix valid API key flagged as invalid when validation request is interrupted (#921)

### LINKED ISSUE #921: API key is displayed as invalid while it's valid in certain scenarios
- Made sure you’re on the latest version => 2.2.3.1

**Describe the bug**
API key isnot accepted although it is correct if we click anywhere on screen while validating it

**To Reproduce**
Scenario1:
1. fresh install of imagify
2. activate the plugin
3. add key and click connect then click anywhere on screen
4. once done, refresh and hover imagify in admin bar => account data not displayed
5.open imagify settings  => key isnot valid displayed

Scenario2:
1- fresh install and activation
2- use valid key
3- edit the key in settings page to be invalid and save 
4- use the right key and click anywhere outside the field
5- save the right key

**Screenshots**
accept valid key even if we clicked anywhere on screen

**Additional context**
- clear the field and save then put the right key and save without clicking anywhere fixes the problem

Miraeld:
# Description

Fixes #921

Users could enter a valid, working API key and still see it flagged as invalid in the UI (admin bar account data missing, settings page showing "Your API key isn't valid!"), even though the key was actually saved. This impacts the very first onboarding step of the plugin, so a false negative here effectively locks users out of the settings page until they stumble on the workaround (re-save the key without clicking elsewhere).

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Manual testing only (see "How to test" below). No JS unit test runner is configured for this plugin, and PHP unit coverage for AJAX admin-post callbacks isn't part of the existing `Tests/Unit` suite conventions (no comparable test exists for sibling callbacks in `class-imagify-admin-ajax-post.php`), so no new automated test was added for the PHP change. `composer test-unit`, `composer run-stan`, and `vendor/bin/phpcs` all pass on the changed file.

### How to test

Reproduction steps from the original issue (both should now succeed instead of showing a false "invalid key" state):

1. On a fresh install, enter your API key and click "connect" (blur the field to trigger the async check), then click anywhere else on the screen while the validation is running. Refresh the page — the key should be stored and treated as valid (admin bar shows account data, settings page shows the "valid" state).
2. With a valid key already saved, edit the field to an invalid key and save. Then re-enter the correct key and click elsewhere (blur) to trigger validation, then save. The key should still be correctly flagged as valid.

Additional regression check: while the "connect" spinner is showing, immediately navigate to another admin page (forcing the browser to abort the in-flight request) — after navigating back to the settings page, the key should still show as valid (this is the scenario the fix specifically targets).

- Imagify settings page: API key field, "connect"/blur validation flow (`assets/js/options.js`, AJAX action `imagify_check_api_key_validity`).
- API key validity caching (`imagify_check_licence_1` site transient) and the "wrong-api-key" admin notice.
- Admin bar account data display (indirectly, since it depends on the API key being correctly flagged as valid).

### Documentation

**Root cause:** `Imagify_Admin_Ajax_Post::imagify_check_api_key_validity_callback()` (`inc/classes/class-imagify-admin-ajax-post.php`) performs two sequential outbound HTTP calls to the Imagify API for a single "check key" request:

1. `get_imagify_status( $api_key )` to validate the key.
2. `update_imagify_option( 'api_key', $api_key )`, which triggers WordPress's `update_option_{option}` hook wired to `Imagify_Settings::after_save_options()`. That method calls `Imagify_Requirements::is_api_key_valid( true )`, which forces a **second**, fresh network call (`get_imagify_user()`) to set the `imagify_check_licence_1` transient and dismiss/renew the "wrong-api-key" notice.

By default, PHP aborts script execution when the client disconnects (the browser cancels the XHR — which happens when the user clicks a link/navigates away while the spinner is showing, per the reported scenarios). If the disconnect happens after step 1 has confirmed the key is valid but while step 2's *internal* validity re-check is still in flight, execution stops before the transient/notice update completes. The net result: the option is saved with the new (valid) key, but the "invalid" notice/lack of cached validity persists until something else happens to refresh it — matching both scenarios in the issue, and explaining why the manual workaround (re-save without clicking away) works: it never lets the browser disconnect mid-request.

**Fix:**
- `inc/classes/class-imagify-admin-ajax-post.php`: wrap the option save + validity refresh in `ignore_user_abort( true )` (restored afterwards) so that, once the key has been confirmed valid, saving it and updating the validity cache/notice always completes as one atomic unit, regardless of whether the client disconnects.
- `assets/js/options.js`: add `.always()` to the AJAX check so the local `busy` lock is always released (including on abort/failure), preventing the blur handler from getting stuck if a request is cancelled. Rebuilt `assets/js/options.min.js` via `npm run build` (Grunt/uglify).

### New dependencies

None.

### Risks

- `ignore_user_abort( true )` is scoped tightly around the option save + validity check and restored to its previous value immediately after, so it has no effect on the rest of the request lifecycle or other AJAX handlers.
- The extra few hundred milliseconds where the script can no longer be killed by a client disconnect only applies to this one endpoint, and only for the duration of two already-existing, bounded outbound HTTP calls — no new long-running work was introduced.

# Mandatory Checklist

## Code validation

## Code style

No automated test was added: there is no JS unit test runner configured in this project, and no existing PHP unit test convention covers the admin-post AJAX callbacks in `class-imagify-admin-ajax-post.php` (they depend on live/mocked outbound HTTP calls and WP AJAX plumbing not currently exercised by `Tests/Unit`). Verification is manual, per the "How to test" section above, following the exact reproduction steps from the issue.

# Additional Checks


remyperona (approved):

Miraeld:
Works as expected QA 🟢

PR #1193, opened 2026-07-21
Enhancement #516: Add imagify_bulk_optimize_media filter to exclude media from bulk optimization

### LINKED ISSUE #516: Add a filter to exclude file types when running bulp optimization
We currently have a filter `imagify_auto_optimize_attachment` to disable certain file types from being optimized automatically in the upload process. Ideally we should have a filter for the bulk optimization process as well.

Miraeld:
# Description

Fixes #516

Developers can now exclude specific media (typically file types like PDFs) from the **bulk optimization** process, mirroring the existing `imagify_auto_optimize_attachment` filter that already covers the automatic on-upload path. This is the most-requested use case in the issue thread: skipping PDFs during bulk runs so their accessibility (508/ADA) features and large originals aren't touched.

A new filter is introduced:

```php
/**
 * @param bool   $optimize           True to optimize this media, false to exclude it.
 * @param int    $media_id           The media ID (attachment ID, or custom-folder file ID).
 * @param string $context            The optimization context ('wp' or 'custom-folders').
 * @param int    $optimization_level The optimization level.
 */
apply_filters( 'imagify_bulk_optimize_media', true, $media_id, $context, $optimization_level );
```

Example — skip PDFs from bulk optimization:

```php
add_filter( 'imagify_bulk_optimize_media', function ( $optimize, $media_id, $context ) {
    if ( 'wp' === $context && 'application/pdf' === get_post_mime_type( $media_id ) ) {
        return false;
    }
    return $optimize;
}, 10, 3 );
```

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated — new unit test `Tests/Unit/classes/Bulk/RunOptimizeTest.php`:
- `testExcludesMediaIdVetoedByFilter` — stub returns IDs `[1,2,3]`, filter vetoes `2`; asserts only `1` and `3` are enqueued and the running transient is set with `total: 2, remaining: 2`.
- `testReturnsNoImagesWhenFilterExcludesAllMedia` — filter vetoes everything → `no-images` result, nothing enqueued.

`composer run test-unit -- --filter RunOptimizeTest` → **OK (2 tests, 10 assertions)**. Regression `--group BulkRunOptimize,BulkRunRestore` → OK (6 tests, 18 assertions). `phpcs` clean on `Bulk.php`.

### How to test

1. Add the example snippet above to a theme/plugin.
2. Upload a PDF and some images, ensuring they are eligible for optimization.
3. Run **Bulk Optimization**. The PDF is excluded from the run; the progress total reflects the exclusion; images optimize normally.
4. Without the filter, behavior is unchanged (PDFs/media still processed as before).

Bulk optimization (`Imagify\Bulk\Bulk::run_optimize()`), both `wp` (media library) and `custom-folders` contexts. Automatic on-upload optimization is unaffected.

### Documentation

The filter is applied once per media ID inside `run_optimize()`, immediately after `get_unoptimized_media_ids()` builds the eligible list and before the emptiness check and enqueue loop. It is purely subtractive — a developer can only remove already-eligible items, never add ineligible ones — exactly like `imagify_auto_optimize_attachment`. Because it runs before `count( $media_ids )` is used for the running transient, the total/remaining progress counters stay correct. The single insertion point covers both contexts, since both flow through `run_optimize()`.

### New dependencies

None.

### Risks

None by default — the filter defaults to `true`, so behavior is unchanged unless a developer opts in. A misbehaving callback can only reduce the work list, never corrupt counts (they are derived after filtering).

# Mandatory Checklist

## Code validation

- [x] I validated all the Acceptance Criteria.

## Code style

N/A — all mandatory items addressed.

# Additional Checks

### DIFF
--- a/Tests/Unit/classes/Bulk/RunOptimizeTest.php
+++ b/Tests/Unit/classes/Bulk/RunOptimizeTest.php
@@ -6,8 +6,8 @@ namespace Imagify\Tests\Unit\classes\Bulk;
 use Brain\Monkey\Filters;
 use Brain\Monkey\Functions;
 use Imagify\Bulk\Bulk;
-use Imagify\Bulk\BulkInterface;
 use Imagify\Tests\Unit\TestCase;
+use Imagify\Tests\Unit\classes\Bulk\Stubs\BulkWithThreeOptimizeIdsStub;
 use Mockery;
 
 /**
@@ -62,14 +62,16 @@ class RunOptimizeTest extends TestCase {
 
 		Functions\expect( 'as_enqueue_async_action' )
 			->twice()
-			->withArgs(
-				function ( $hook, $args, $group ) {
-					return 'imagify_optimize_media' === $hook
-						&& in_array( $args['id'], [ 1, 3 ], true )
-						&& 'wp' === $args['context']
-						&& 1 === $args['level']
-						&& 'imagify-wp-optimize-media' === $group;
-				}
+			->with(
+				'imagify_optimize_media',
+				Mockery::on(
+					function ( $args ) {
+						return in_array( $args['id'], [ 1, 3 ], true )
+							&& 'wp' === $args['context']
+							&& 1 === $args['level'];
+					}
+				),
+				'imagify-wp-optimize-media'
 			);
 
 		Functions\expect( 'set_transient' )
@@ -122,52 +124,5 @@ class RunOptimizeTest extends TestCase {
 			],
 			$result
 		);
-	}
-}
-
-/**
- * Stub bulk: three eligible media IDs for optimization.
- */
-class BulkWithThreeOptimizeIdsStub implements BulkInterface {
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_unoptimized_media_ids( $optimization_level ) {
-		return [ 1, 2, 3 ];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_optimized_media_ids(): array {
-		return [];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_optimized_media_ids_without_format( $format ) {
-		return [
-			'ids'    => [],
-			'errors' => [
-				'no_file_path' => [],
-				'no_backup'    => [],
-			],
-		];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function has_optimized_media_without_nextgen() {
-		return 0;
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_context_data() {
-		return [];
 	}
 }
--- a/Tests/Unit/classes/Bulk/RunOptimizeTest.php
+++ b/Tests/Unit/classes/Bulk/RunOptimizeTest.php
@@ -1,0 +1,173 @@
+<?php
+declare(strict_types=1);
+
+namespace Imagify\Tests\Unit\classes\Bulk;
+
+use Brain\Monkey\Filters;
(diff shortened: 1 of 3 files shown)


classes/Bulk/Bulk.php
```diff
@@ -183,6 +183,28 @@ public function run_optimize( string $context, int $optimization_level ): array
...
+		 * Return false to exclude a given media from the bulk optimization run
+		 * (e.g. to skip a specific file type such as PDFs).
+		 *
+		 * @since 2.3
+		 *
+		 * @param bool   $optimize           True to optimize this media, false to exclude it.
+		 * @param int    $media_id           The media ID (attachment ID, or custom-folder file ID).
+		 * @param string $context            The optimization context ('wp' or 'custom-folders').
+		 * @param int    $optimization_level The optimization level.
+		 */
```
remyperona: This should be right next to the filter itself

classes/Bulk/Bulk.php
```diff
@@ -183,6 +183,28 @@ public function run_optimize( string $context, int $optimization_level ): array
...
+		$media_ids = array_values(
+			array_filter(
+				$media_ids,
+				function ( $media_id ) use ( $context, $optimization_level ) {
+					return apply_filters( 'imagify_bulk_optimize_media', true, $media_id, $context, $optimization_level );
```
remyperona: PHPStan is reporting the discouraged usage, might be worth to switch to the typed function

Tests/Unit/classes/Bulk/RunOptimizeTest.php
```diff
@@ -0,0 +1,173 @@
...
+			->times( 3 )
+			->andReturnUsing(
+				function ( $optimize, $media_id ) {
+					return 2 !== $media_id;
+				}
+			);
+
+		Functions\expect( 'as_enqueue_async_action' )
+			->twice()
+			->withArgs(
```
remyperona: Did you check this report from PHPStan?

Tests/Unit/classes/Bulk/RunOptimizeTest.php
```diff
@@ -0,0 +1,173 @@
...
+			],
+			$result
+		);
+	}
+}
+
+/**
+ * Stub bulk: three eligible media IDs for optimization.
+ */
+class BulkWithThreeOptimizeIdsStub implements BulkInterface {
```
remyperona: stub class should have its own file

remyperona (changes requested):

classes/Bulk/Bulk.php
Miraeld: Moved it inside the closure, right above the filter call

classes/Bulk/Bulk.php
Miraeld: Switched to wpm_apply_filters_typed('boolean', …)

Tests/Unit/classes/Bulk/RunOptimizeTest.php
Miraeld: change `withArgs()` (undefined per PHPStan) → `->with() + Mockery::on()`,  the repo's convention

Tests/Unit/classes/Bulk/RunOptimizeTest.php
Miraeld: Extracted `BulkWithThreeOptimizeIdsStub` to `Tests/Unit/classes/Bulk/Stubs/`

remyperona (approved):

sandyfzu (approved):

PR #1194, opened 2026-07-21
Fix #1041: Fix indefinite spinner and stuck progress in Generate Missing Next-Gen

### LINKED ISSUE #1041: Generate Missing Next-Gen Images: Fix indefinite spinner and stuck progress counter
## Summary

This report documents a cluster of bugs that collectively caused the "Generate Missing Next-Gen Images Version" bulk operation to spin indefinitely and never make progress. The root causes span SQL query correctness, early-abort logic, and a data storage key corruption bug. All bugs were found by tracing a real-world reproduction where the progress counter showed `0/21` (or similar) and never advanced, even after page refresh.

Seven distinct bugs are fixed across four files.

---

## Background

The bulk next-gen generation flow works roughly as follows:

1. A transient is set with the total count of images needing next-gen versions (`imagify_missing_next_gen_total`)
2. `run_generate_nextgen()` calls `get_optimized_media_ids_without_format()` on each context to collect IDs
3. Those IDs are dispatched to ActionScheduler
4. Each AS job calls `generate_nextgen_versions()`, which reads existing optimization data and writes the webp result back to the database
5. The UI polls progress by re-counting how many images still lack the format key

If any step silently fails — IDs that shouldn't be queued get queued, AS marks jobs complete but nothing is written, the NOT LIKE query keeps returning the same rows — the counter never moves.

---

### Bug 1 — `classes/Bulk/CustomFolders.php`: Inactive folder files included in count

**Method:** `get_optimized_media_ids_without_format()`

Files belonging to inactive custom folders were counted in the "remaining" total but were never enqueued for processing. The SQL JOIN to the folders table did not filter on `fo.active = 1`, so inactive-folder files inflated the count with entries that could never be processed.

```diff
- INNER JOIN $folders_table AS fo
-     ON ( fi.folder_id = fo.folder_id )
+ INNER JOIN $folders_table AS fo
+     ON ( fi.folder_id = fo.folder_id AND fo.active = 1 )
```

---

### Bug 2 — `classes/Bulk/CustomFolders.php`: NULL data rows incorrectly included

**Method:** `get_optimized_media_ids_without_format()`

The WHERE clause included `OR fi.data IS NULL`, which admitted files with no optimization data into the result set. `generate_nextgen_versions()` requires existing optimization data to operate; when called on these files it returns early without writing anything. ActionScheduler marks the job complete, but no `@imagify-webp` key is ever written, so the same files are returned on every subsequent poll.

```diff
- AND ( fi.data NOT LIKE %s OR fi.data IS NULL )
+ AND fi.data NOT LIKE %s
```

---

### Bug 3 — `classes/Bulk/Bulk.php`: `no-backup` error aborted entire operation

**Method:** `run_generate_nextgen()`

When iterating over multiple contexts (e.g., `custom-folders` followed by `wp`), a `no-backup` error on the first context triggered an immediate `return ['success' => false, 'message' => 'no-backup']`, aborting the loop before subsequent contexts were ever evaluated. Users with custom folders that had no backups would silently get zero processing done on their media library — despite the media library images being perfectly valid candidates.

```diff
  foreach ( $contexts as $context ) {
      foreach ( $formats as $format ) {
          $media = $this->get_bulk_instance( $context )->get_optimized_media_ids_without_format( $format );
-         if ( ! $media['ids'] && $media['errors']['no_backup'] ) {
-             return [ 'success' => false, 'message' => 'no-backup' ];
-         } elseif ( ! $media['ids'] && $media['errors']['no_file_path'] ) {
+         if ( ! $media['ids'] && $media['errors']['no_file_path'] ) {
              return [ 'success' => false, 'message' => __( 'The path to the backup file...', 'imagify' ) ];
          }
-         $medias[ $context ] = $media['ids'];
+         if ( $media['ids'] ) {
+             $medias[ $context ] = $media['ids'];
+         }
      }
  }
```

---

### Bug 4 — `classes/Bulk/WP.php`: Unoptimized images included via NULL status join

**Method:** `get_optimized_media_ids_without_format()`

The SQL condition `AND (mt1.meta_key IS NULL OR mt1.meta_value = 'success' OR mt1.meta_value = 'already_optimized')` inadvertently included images that had `_imagify_data` stored but no corresponding `_imagify_status` row. These images fail inside `generate_nextgen_versions()` with `WP_Error('not_optimized')`. AS marks the job complete, nothing is written, and the count does not move.

```diff
- AND (mt1.meta_key IS NULL OR mt1.meta_value = 'success' OR mt1.meta_value = 'already_optimized')
+ AND (mt1.meta_value = 'success' OR mt1.meta_value = 'already_optimized')
```

---

### Bug 5 — `classes/Optimization/Process/AbstractProcess.php`: WebP size key corrupted when API reports "WebP larger than original"

**Method:** `update_size_optimization_data()` (~line 1973)

This is the most impactful bug and the primary root cause of the indefinite spinner for most users.

When the Imagify API returns a response indicating that the WebP version would be larger than the original (the response object carries a `message` property), the code executed:

```php
if ( property_exists( $response, 'message' ) ) {
    $size = str_replace( $this->format, '', $size );
}
$this->get_data()->update_size_optimization_data( $size, $data );
```

For a size named `full@imagify-webp`, the `str_replace` produces `full`. The result is stored under the `full` key, overwriting the original optimization data for that size. The `full@imagify-webp` key is **never written**.

Because the NOT LIKE query checks for the presence of the `@imagify-webp` key with `success;b:1`, these images match forever. Every click of "Generate Missing Next-Gen" re-queues them, AS processes them, the API again reports "WebP larger than original", the key is again stored under the wrong name, and the counter never reaches zero.

This affects every image where the WebP output would be larger than the source — a routine outcome for already-optimized images, PNGs with transparency, and small files. On many real-world sites this is the majority of the queue.

Storing the result under the correct `@imagify-webp` key is the right behaviour regardless of whether the API chose not to replace the file. The data has `success: true`, satisfies the NOT LIKE check, and the image is correctly excluded from all future runs.

```diff
- if ( property_exists( $response, 'message' ) ) {
-     $size = str_replace( $this->format, '', $size );
- }
  $this->get_data()->update_size_optimization_data( $size, $data );
```

---

### Bug 6 — `classes/Bulk/WP.php` and `classes/Bulk/CustomFolders.php`: Mime type exclusion silently fails due to whitespace

**Method:** `get_optimized_media_ids_without_format()` in both files

The code removes the target format's mime type from the SQL `IN (...)` clause so that existing `.webp` source files are not asked to generate a webp version of themselves. It does this by splitting the string returned by `Imagify_DB::get_mime_types()`, finding the matching entry, then using `str_replace` to remove it.

When `get_mime_types()` returns a comma-space-separated string (e.g., `'image/jpeg', 'image/webp'`), the extracted entry after splitting is `' image/webp'` with a leading space. The `str_replace` then looks for `,' image/webp'` in the original string, but the original uses `,'image/webp'` — no match, and `image/webp` is never removed from the query.

```diff
+ $mime = trim( $mime );
- $mime_types = str_replace( ",'" . $mime . "'", '', $mime_types );
+ $mime_types = str_replace( [ ", '" . $mime . "'", ",'" . $mime . "'" ], '', $mime_types );
```

---

### Bug 7 — `classes/Bulk/WP.php` and `classes/Bulk/CustomFolders.php`: `.webp` files stored with wrong mime type permanently stuck

**Method:** `get_optimized_media_ids_without_format()` in both files (PHP post-processing loop)

Some third-party plugins (BuddyBoss confirmed) upload `.webp` files but register them in `wp_posts` with `post_mime_type = 'image/jpeg'` or another non-webp value. The SQL mime-type exclusion (Bug 6) relies on a correct `post_mime_type`; for these files it has no effect. The file is included in the result set, AS queues and processes it, `generate_nextgen_versions()` finds a `.webp` source and has nothing to do, AS marks the job complete, the `@imagify-webp` key is never written, and the file is permanently "remaining = 1".

Added to both files before the backup existence check:

```diff
+ // Skip files whose extension already matches the target format
+ // (e.g. a .webp file stored with incorrect post_mime_type).
+ if ( strtolower( pathinfo( $file_path, PATHINFO_EXTENSION ) ) === $format ) {
+     continue;
+ }
```

---

## Files Changed

| File | Bugs Fixed |
|---|---|
| `classes/Bulk/CustomFolders.php` | Bug 1, Bug 2, Bug 6, Bug 7 |
| `classes/Bulk/Bulk.php` | Bug 3 |
| `classes/Bulk/WP.php` | Bug 4, Bug 6, Bug 7 |
| `classes/Optimization/Process/AbstractProcess.php` | Bug 5 |

---

## How to Reproduce

**Bug 5 (most common — affects nearly all sites):**
1. Have at least one image optimized by Imagify where the WebP version would be larger than the original
2. Go to **Settings → Imagify** and click **Generate Missing Next-Gen Images Version**
3. The counter shows `N` remaining and never decreases
4. Inspect `_imagify_data` for those images — the `@imagify-webp` key is absent; a spurious entry has been written under the bare size key (e.g., `full`) instead

**Bug 1 + Bug 2 (custom folders):**
1. Configure a custom folder in Imagify settings, then deactivate it
2. Ensure some files in that folder have no `fi.data` in the custom files table
3. Run "Generate Missing Next-Gen" — the total includes these files but they are never processed

**Bug 3 (custom folders + media library):**
1. Configure custom folders with images that have no backup files
2. Run "Generate Missing Next-Gen" — media library images are silently skipped entirely

**Bug 4 (partially-processed media library entries):**
1. Have a media library image with `_imagify_data` present but no `_imagify_status` meta row
2. Run "Generate Missing Next-Gen" — the image is queued and silently fails every time

**Bug 6 + Bug 7 (webp source files):**
1. Upload a `.webp` file via a plugin that stores it with `post_mime_type = 'image/jpeg'`
2. Run "Generate Missing Next-Gen" — the file appears permanently as "remaining = 1"

---

## Investigation Notes

1. Confirmed the AJAX call fired correctly; the `imagify_missing_next_gen_total` transient was set with the expected count
2. Monitored ActionScheduler — jobs were created and completed immediately with no database changes
3. Queried `_imagify_data` for all affected post IDs — none had an `@imagify-webp` key; all had a root-level `message` of "WebP file is larger than the original image"
4. Traced `update_size_optimization_data()` and found the `str_replace` corrupting the storage key (Bug 5)
5. After fixing Bug 5, progress reached 18/19; the remaining image had `post_mime_type = 'image/jpeg'` but a `.webp` file extension (Bug 7)
6. Bugs 1–4 were uncovered while tracing why `0/N` never progressed before any AS jobs ran

Miraeld:
# Description

Fixes #1041

The **Generate Missing Next-Gen Images** bulk operation could spin indefinitely with a progress counter (e.g. `0/21`) that never advanced, even after a page refresh. This PR fixes a cluster of **7 distinct bugs** across 4 files that collectively caused images to be counted as "remaining" but never successfully processed — so the counter never reached zero.

The most impactful is Bug 5: when the Imagify API reported that the WebP version would be larger than the original, the next-gen result was written under the wrong data key, so those images matched the "remaining" query forever and were re-queued on every run. This alone stuck the majority of real-world queues.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

- Automated: new unit test `Tests/Unit/classes/Bulk/RunGenerateNextgenTest.php` (`@group GenerateNextgen`) covering Bug 3 — a context that has only `no_backup` errors (empty ids) no longer aborts the whole run; a second context with valid ids is still scheduled. Full unit suite: **346 tests pass**, no new failures (17 pre-existing risky/deprecation warnings unchanged).
- `php -l` passes on all 4 changed files and the new test.

### How to test

**Bug 5 (most common):**
1. Have at least one image optimized by Imagify where the WebP version would be larger than the original (routine for already-optimized images, transparent PNGs, small files).
2. Settings → Imagify → **Generate Missing Next-Gen Images Version**.
3. Before fix: counter never decreases; `_imagify_data` for those images has no `@imagify-webp` key (a spurious entry is written under the bare size key). After fix: those images get the `@imagify-webp` key with `success: true` and are excluded from future runs; the counter advances to zero.

**Bug 1 + 2 (custom folders):** deactivate a custom folder containing files with no `fi.data`; before fix the total includes files that are never processed.

**Bug 3 (custom folders + media library):** custom-folder images with no backups previously aborted the whole run, silently skipping the media library; now the media library still processes.

**Bug 4 (partial media entries):** a media image with `_imagify_data` but no `_imagify_status` row was queued and silently failed each time.

**Bug 6 + 7 (webp source files):** a `.webp` uploaded with `post_mime_type = image/jpeg` (e.g. BuddyBoss) was permanently stuck at "remaining = 1".

Bulk "Generate Missing Next-Gen Images" for both media library (`wp`) and custom folders; the next-gen size-data storage in the optimization process. Standard image optimization is not changed.

### Documentation

Seven fixes:

| File | Fix |
|---|---|
| `classes/Bulk/CustomFolders.php` | **Bug 1** JOIN now filters `fo.active = 1`; **Bug 2** drop `OR fi.data IS NULL`; **Bug 6** trim mime + position-independent removal from `IN(...)`; **Bug 7** skip files whose extension already matches the target format |
| `classes/Bulk/WP.php` | **Bug 4** drop `mt1.meta_key IS NULL OR` from the status condition; **Bug 6**; **Bug 7** |
| `classes/Bulk/Bulk.php` | **Bug 3** `run_generate_nextgen()` no longer returns early on a `no-backup` context; only assigns `$medias[$context]` when ids exist, so later contexts are still evaluated |
| `classes/Optimization/Process/AbstractProcess.php` | **Bug 5** remove the `str_replace` that rewrote the size key when the API response had a `message`, so next-gen data is always stored under the correct `size@imagify-webp` key |

### New dependencies

None.

### Risks

Low. Bug 5 changes storage to always use the correct key — the data carries `success: true`, so the image is correctly excluded from future runs regardless of whether the API chose to replace the file. Bug 3 relies on the existing downstream `empty( $medias )` guard as the terminal no-images case. SQL changes are narrowing (fewer false candidates), so they cannot cause over-processing.

# Mandatory Checklist

## Code validation

- [x] I validated all the Acceptance Criteria.

## Code style

Unit coverage focuses on Bug 3 (the most unit-testable in isolation). Bugs 1/2/4 are `$wpdb`-query-shape changes and Bug 5 is verified via the reproduction scenario; these are better exercised by integration/manual QA than by mocking `$wpdb`.

# Additional Checks

### DIFF
--- a/Tests/Unit/classes/Bulk/RunGenerateNextgenTest.php
+++ b/Tests/Unit/classes/Bulk/RunGenerateNextgenTest.php
@@ -6,8 +6,9 @@ namespace Imagify\Tests\Unit\classes\Bulk;
 use Brain\Monkey\Filters;
 use Brain\Monkey\Functions;
 use Imagify\Bulk\Bulk;
-use Imagify\Bulk\BulkInterface;
 use Imagify\Tests\Unit\TestCase;
+use Imagify\Tests\Unit\classes\Bulk\Stubs\BulkNoBackupOnlyStub;
+use Imagify\Tests\Unit\classes\Bulk\Stubs\BulkWithIdsStub;
 use Mockery;
 
 /**
@@ -64,99 +65,5 @@ class RunGenerateNextgenTest extends TestCase {
 			],
 			$result
 		);
-	}
-}
-
-/**
- * Stub bulk: only `no_backup` errors, no eligible ids and no `no_file_path` errors.
- */
-class BulkNoBackupOnlyStub implements BulkInterface {
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_unoptimized_media_ids( $optimization_level ) {
-		return [];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_optimized_media_ids(): array {
-		return [];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_optimized_media_ids_without_format( $format ) {
-		return [
-			'ids'    => [],
-			'errors' => [
-				'no_file_path' => [],
-				'no_backup'    => [ 1, 2 ],
-			],
-		];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function has_optimized_media_without_nextgen() {
-		return 0;
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_context_data() {
-		return [];
-	}
-}
-
-/**
- * Stub bulk: two eligible media ids, no errors.
- */
-class BulkWithIdsStub implements BulkInterface {
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_unoptimized_media_ids( $optimization_level ) {
-		return [];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_optimized_media_ids(): array {
-		return [];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_optimized_media_ids_without_format( $format ) {
-		return [
-			'ids'    => [ 10, 20 ],
-			'errors' => [
-				'no_file_path' => [],
-				'no_backup'    => [],
-			],
-		];
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function has_optimized_media_without_nextgen() {
-		return 0;
-	}
-
-	/**
-	 * {@inheritdoc}
-	 */
-	public function get_context_data() {
-		return [];
 	}
 }
--- a/Tests/Unit/classes/Bulk/RunGenerateNextgenTest.php
+++ b/Tests/Unit/classes/Bulk/RunGenerateNextgenTest.php
@@ -1,0 +1,162 @@
+<?php
+declare(strict_types=1);
+
+namespace Imagify\Tests\Unit\classes\Bulk;
+
+use Brain\Monkey\Filters;
+use Brain\Monkey\Functions;
+use Imagify\Bulk\Bulk;
+use Imagify\Bulk\BulkInterface;
+use Imagify\Tests\Unit\TestCase;
+use Mockery;
+
+/**
+ * Tests for \Imagify\Bulk\Bulk::run_generate_nextgen().
+ *
+ * `Imagify_Requirements` is a legacy classmap-autoloaded class; it is stubbed via
+ * a Mockery alias mock, so this test runs in its own process to guarantee the
+ * alias is registered before the real class would otherwise be autoloaded.
+ *
+ * @covers \Imagify\Bulk\Bulk::run_generate_nextgen
+ * @group  GenerateNextgen
+ * @since  2.3
+ *
+ * @runTestsInSeparateProcesses
+ * @preserveGlobalState disabled
+ */
+class RunGenerateNextgenTest extends TestCase {
+
+	/**
+	 * Stubs `Imagify_Requirements::is_api_key_valid()` / `is_over_quota()` via a Mockery
+	 * alias mock so `can_optimize()` lets execution proceed past the initial guard.
+	 */
+	private function stubRequirements(): void {
+		$mock = Mockery::mock( 'alias:Imagify_Requirements' );
+		$mock->shouldReceive( 'is_api_key_valid' )->andReturn( true );
+		$mock->shouldReceive( 'is_over_quota' )->andReturn( false );
+	}
+
+	/**
+	 * Regression guard: a context that only has `no_backup` errors (empty ids) must not
+	 * abort the whole run. Another context with valid ids must still be scheduled.
+	 */
+	public function testContextWithOnlyNoBackupErrorsDoesNotAbortOtherContexts(): void {
+		$this->stubRequirements();
+
+		Functions\when( 'delete_transient' )->justReturn( true );
+		Functions\when( 'set_transient' )->justReturn( true );
+		Functions\when( 'as_enqueue_async_action' )->justReturn( 1 );
+
+		Filters\expectApplied( 'imagify_bulk_class_name' )
+			->twice()
+			->andReturnUsing(
+				function ( $class_name, $context ) {
+					return 'wp' === $context ? BulkNoBackupOnlyStub::class : BulkWithIdsStub::class;
+				}
+			);
+
+		$result = ( new Bulk() )->run_generate_nextgen( [ 'wp', 'custom-folders' ], [ 'webp' ] );
+
+		$this->assertSame(
+			[
+				'success' => true,
+				'message' => 2,
+			],
+			$result
+		);
+	}
+}
+
+/**
+ * Stub bulk: only `no_backup` errors, no eligible ids and no `no_file_path` errors.
+ */
+class BulkNoBackupOnlyStub implements BulkInterface {
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_unoptimized_media_ids( $optimization_level ) {
+		return [];
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_optimized_media_ids(): array {
+		return [];
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_optimized_media_ids_without_format( $format ) {
+		return [
+			'ids'    => [],
+			'errors' => [
+				'no_file_path' => [],
+				'no_backup'    => [ 1, 2 ],
+			],
+		];
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function has_optimized_media_without_nextgen() {
+		return 0;
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_context_data() {
+		return [];
+	}
+}
+
+/**
+ * Stub bulk: two eligible media ids, no errors.
+ */
+class BulkWithIdsStub implements BulkInterface {
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_unoptimized_media_ids( $optimization_level ) {
+		return [];
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_optimized_media_ids(): array {
+		return [];
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function get_optimized_media_ids_without_format( $format ) {
+		return [
+			'ids'    => [ 10, 20 ],
+			'errors' => [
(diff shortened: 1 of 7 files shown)


Tests/Unit/classes/Bulk/RunGenerateNextgenTest.php
```diff
@@ -0,0 +1,162 @@
...
+			],
+			$result
+		);
+	}
+}
+
+/**
+ * Stub bulk: only `no_backup` errors, no eligible ids and no `no_file_path` errors.
+ */
+class BulkNoBackupOnlyStub implements BulkInterface {
```
remyperona: This should be split in its own class

remyperona (changes requested):

remyperona (approved):

Miraeld:
## Manual QA — confirmed working ✅

Tested against this PR's exact commit (`51d3ec15`) on a local WP install.

**Bug 5 (highest-impact fix) — deterministic verification**

Rather than trying to organically coax the Imagify API into a "WebP would be bigger than original" response, called `AbstractProcess::update_size_optimization_data()` directly (via `wp eval-file`) on a real, already-optimized attachment, with a fake API response object carrying a `message` property — the exact condition the old `str_replace` branch keyed off of:

```php
$response->message = 'Original file is smaller than the optimized one.';
$process->update_size_optimization_data( $response, 'full@imagify-webp', 'normal' );
```

Result — `_imagify_data.sizes` keys after the call:
```
full, medium_large@imagify-webp, thumbnail@imagify-webp, medium@imagify-webp,
full@imagify-webp, medium, thumbnail, medium_large
```

`full@imagify-webp` is stored as its own distinct key alongside the plain `full` (regular optimization data) — no collision, no stripped suffix. Before this fix, the `str_replace( $this->format, '', $size )` branch would have collapsed `full@imagify-webp` down to `full`, overwriting/colliding with the non-webp entry and permanently matching the "remaining" SQL query in `get_optimized_media_ids_without_format()` — this is the mechanism that caused the counter to never reach zero.

**Bugs 1–4, 6, 7 (SQL/query fixes in `CustomFolders.php`, `WP.php`, `Bulk.php`)**

These are narrowing SQL/query-shape changes and a control-flow fix (no longer aborting the whole run on a `no_backup`-only context) — reviewed against the diff and covered by the new `RunGenerateNextgenTest` unit test (`no_backup`-only context no longer blocks other contexts from being scheduled). Not independently reproduced end-to-end in this pass (would require deliberately corrupting `post_mime_type`/custom-folder data to match each specific edge case), but the logic checks out and risk is low per the PR's own analysis (queries only get more restrictive, never less).

**Result:** the core fix (Bug 5) is verified working with a direct, reproducible test. Recommend merging; the remaining bugs are lower-risk query narrowing already covered by unit tests.

PR #1196, opened 2026-07-23
fix: ajaxify the attachment edition page buttons

faisalahammad:
# Description

Ajaxify the Optimize / Re-optimize / Restore buttons on the Edit Media page so clicks no longer trigger a full plugin reload through `admin-post.php`. The page stays put, the "Optimizing…" / "Restoring…" spinner replaces the button in place, and the Imagifybeat polling path takes over when the work is queued in the background. No-JS fallback unchanged.

Fixes #434

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### What was tested
- New unoptimized attachment (jpg)
- Already-optimized attachment (with backup) for Restore / Re-optimize
- Attachment in locked state (server-side work already running)
- No-JS fallback path

### How to test
1. Media > Library > click an attachment > Edit Media page.
2. Click Optimize. Expect: no page reload, in-place spinner, AJAX request to `admin-ajax.php?action=imagify_manual_optimize`, panel updates.
3. Click Restore then Re-optimize on optimized item. Same in-place flow.
4. Re-load an attachment whose process is locked server-side. Spinner clears within ~15s via Imagifybeat.

### Affected Features & Quality Assurance Scope
- Attachment Edit Media (`post.php?action=edit`)
- Imagifybeat AJAX polling
- Imagify media modal (library page)

### Documentation
No doc change required; user-facing behaviour fix only.

### New dependencies
None.

### Risks
- The `imagify-media-modal` script is now enqueued on the attachment edit screen in addition to the library page. No double-enqueue risk because the library already loads it via dependency.
- If `class_exists()` / `interface_exists()` autoload fallback fires (non-Strauss install), Composer autoload is preserved.

# Mandatory Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Lint and static analysis pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1197, opened 2026-07-23
test(tracking): add integration test for imagify_after_optimize hook

### LINKED ISSUE #1159: Add integration test for imagify_after_optimize hook

**Source:** Follow-up from QA on #1126 (COULD_HAVE)

**Context**
The new `classes/Tracking/Subscriber` registers on `imagify_after_optimize`, but there is no integration test verifying the full hook-to-event-fire chain end-to-end. Currently only unit tests with mocked dependencies cover the tracking logic.

**Suggestion**
Add an integration test that verifies the `imagify_after_optimize` hook is actually registered and wired to the Subscriber, with the full stack exercised (not just mocked dependencies).

**Acceptance Criteria**
- [ ] Integration test exists that fires `imagify_after_optimize` and verifies the Subscriber is invoked
- [ ] Test covers the full hook-to-event chain without mocking the Subscriber
- [ ] Test passes in CI

**Development steps**
- [ ] Create integration test file in appropriate test directory
- [ ] Set up test fixture with real Subscriber instance
- [ ] Fire `imagify_after_optimize` hook with test data
- [ ] Assert Subscriber callback was executed

**Effort estimation**
S / M

faisalahammad:
## Description
Integration test that fires the `imagify_after_optimize` hook and verifies the Tracking Subscriber callback runs end to end. Resolves the gap where existing tests only asserted registration via `has_action()` without exercising the full hook-to-callback chain.

## Summary
Adds an integration test for the `imagify_after_optimize` hook that verifies the Tracking Subscriber is invoked through the full event chain. The existing `Test_GetSubscribers` test only checked that the Subscriber was registered via `has_action()`, but never fired the action to confirm the callback actually runs end to end.

Closes #1159

### Tests/Integration/classes/Tracking/Subscriber/Test_OptimizeHookIntegration.php
**Before:**
No test fired `imagify_after_optimize`. Subscriber registration was asserted statically only.
**After:**
```php
public function set_up(): void {
    parent::set_up();
    $container   = apply_filters( imagify_container, null );
    $subscriber  = $container->get( Subscriber::class );
    $this->mock_tracking = $this->createMock( Tracking::class );

    $tracking_property = new \ReflectionProperty( Subscriber::class, tracking );
    $tracking_property->setAccessible( true );
    $tracking_property->setValue( $subscriber, $this->mock_tracking );
}

public function testFiringImagifyAfterOptimizeCallsTrackMediaOptimized(): void {
    // ...
    $this->mock_tracking->expects( $this->once() )
        ->method( track_media_optimized )
        ->with( $mock_process, $item );

    do_action( imagify_after_optimize, $mock_process, $item );
}
```
**Why:** Resolves the real Subscriber from the DI container and swaps only the Tracking dependency for a mock, so the Subscriber itself is not mocked. Firing `do_action()` exercises the complete path: ServiceProvider registration, EventManager wiring, and the Subscriber callback. The PHPUnit mock expectation auto-verifies the delegation.

## Testing
**Test 1: Hook fires and Subscriber callback runs**
1. Run `composer test-integration -- --filter="Test_OptimizeHookIntegration"`
Result: test passes; `track_media_optimized()` is called once with the expected process and item.


Miraeld:
Thanks @faisalahammad - picking this up. Your diagnosis and the listener-side test were the right starting point, and both carry over.

Continued in #1258 rather than pushing here, for two reasons:

**1. This branch was cut from #1196, not from `develop`.** So it carries #1196's 10 non-test files (`inc/admin/meta-boxes.php`, `inc/classes/class-imagify-assets.php`, `inc/main.php`, the e2e specs) as scope creep on a test-only ticket. `develop` has also since landed the `ReflectionProperty::setValue()` PHP 8.4 fix and the e2e API-key skip guards independently, and the branch is 63 commits behind. Reconciling that needed a merge plus reverts, which would have meant force-pushing over your fork - not something to do to someone else's branch.

**2. The test needed to cover the producer.** It fired `do_action( 'imagify_after_optimize', ... )` itself, which exercises the Subscriber wiring but not the production call site. I checked by deleting the `do_action()` in `MediaOptimization::task_after()`: all **436 unit + integration tests stayed green**, this one included. So as a regression guard for #1159 it was inert.

#1258 keeps your listener test and adds two producer-contract tests driving the real `task_after()`. Deleting either `do_action()` now fails the matching test.

Also for reference: the `task-check` failure here was just the PR body missing the `## Type of change` section from the template - worth knowing for #1196, which is still open and useful on its own.

PR #1201, opened 2026-07-26
Closes #541: Add imagify_site_root_url filter for domain-mapped multisites

### LINKED ISSUE #541: Improve get_site_root_url() with additional filter for multisites (with domain mapping)
We have user who uses domain mapping for their multisite. By enabling the automatic WebP display option using **picture**  tag, they noticed that only the URL that matches the multisite’s base URL gets updated with picture tags. All other domains don’t.

From user's feedback:

> After digging deeper I could find the method get_site_root_url() in the class Imagify_Filesystem where the domain of the current network is hardcoded as base so that every domain not matching this base URL cannot be detected successfully as an internal domain.

> Thus, I would like to get a filter here in a future version of Imagify to allow me to change this domain as needed.

And one more thing that might be improved:

> Additionally, setting `$root_url = is_ssl() ? 'https' : 'http';` in this method doesn’t make sense since you immediately overwrite this variable with the base URL of the network.

Miraeld:
# Description

Closes #541

On multisites using domain mapping, `get_site_root_url()` returns the network's own domain/path. For mapped subdomains, this never matches incoming request URLs, causing the next-gen `<picture>` rewrite to silently do nothing for non-uploads image URLs. This PR introduces an `imagify_site_root_url` filter to allow integrators to specify the correct URL for their domain-mapped subsite without changing defaults.

## Type of change

- [ ] Sub-task of #541
- [ ] Chore
- [ ] Release

### What was tested

**Automated tests (unit):**
- `composer test-unit -- --filter="GetSiteRootUrl"` → 16 tests, 34 assertions, all passing
- `composer test-unit` (full suite) → 371 tests, 867 assertions, 17 risky, 0 failures. Baseline on this branch before these tests was 355 tests / 833 assertions / 17 risky, so no regressions were introduced.
- `composer phpcs inc/classes/class-imagify-filesystem.php` → 0 violations

**Static analysis:**
- `composer run-stan` → **2 errors, both pre-existing and unrelated** (`classes/Media/WP.php:214`, `requireOnce.fileNotFound`). Verified present on the branch point too. This PR neither introduces nor fixes them.

**Integration tests (unrun):**
An integration suite for `GetSiteRootUrl` is committed but has not been executed — there is no WordPress test installation in the authoring environment. Note that the CI code-coverage job does not install WordPress or MySQL either, so these tests do not currently run anywhere; they are covered by the unit suite instead (see below).

**Manual domain-mapping verification (not performed):**
The actual domain-mapping fix was not manually verified — it requires a real domain-mapped multisite setup, which was unavailable. This is the **most critical QA item before merge**: on a mapped multisite with next-gen enabled, hook `imagify_site_root_url` to return `home_url( '/' )` for the mapped subsite and confirm non-uploads image URLs on the mapped domain get rewritten; with no filter hooked, confirm behaviour is byte-identical to the current release.

### How to test

1. **Unit tests:** Run `composer test-unit` to verify core logic, including the full `get_site_root_url()` behaviour.
2. **Manual domain mapping (required before merge):** On a real WordPress multisite with domain mapping enabled:
   - Enable next-gen image display in Imagify settings
   - Add a callback to the `imagify_site_root_url` filter that returns the correct URL for your mapped subsite
   - Verify that non-uploads image URLs on the mapped domain now generate `<picture>` tags
   - Remove the callback and verify next-gen display reverts to the previous (broken) state
3. **Regression:** Verify that without the filter hooked, existing single-site and multisite main-site behaviour is unchanged.

- **Next-gen `<picture>` display and URL matching:** The primary change is in how `get_site_root_url()` caches and exposes its value; downstream code in `classes/Picture/Display.php:733` (`url_to_path()`) uses this to determine whether to rewrite image URLs.
- **Custom-folder `{{ROOT}}` placeholder resolution:** `inc/classes/class-imagify-files-scan.php:484` also consumes `get_site_root_url()` for template expansion.
- **WP Rocket integration:** `inc/3rd-party/wp-rocket/classes/Main.php:78` uses `get_site_root_url()` for scheme detection.

All three code paths are affected by the caching change and should be regression-tested.

### How it works

The old implementation used a function-static variable `static $root_url` to cache the computed site root URL. This cache is process-wide and not keyed by blog, so:

1. On multisite, when switching blogs via `switch_to_blog()` (WP-CLI network bulk runs, network-admin flows), the function would return the first computed value for all subsequent blog contexts.
2. In tests, the static cache couldn't be reset, making the filter untestable.

This PR replaces the function-static with a blog-keyed instance property on the existing `Imagify_Filesystem` singleton. The method is restructured to:

1. Check if the value is already cached for the current blog.
2. If not, compute it the same way as before.
3. **Fire the `imagify_site_root_url` filter** before caching (allowing integrators to override).
4. Validate the filtered value: it is used only if it is a non-blank string, and is then trimmed and passed through `trailingslashit()`. Anything else — array, object, integer, `null`, `false`, `0`, empty or whitespace-only string — is discarded in favour of `home_url( '/' )`.
5. Cache the value keyed by blog ID.

The filter fires on all code paths (single-site, multisite main site, and multisite subsite), which is deliberate: a filter that fires on only one branch is a support trap.

**Default behaviour is unchanged** — when no callback is hooked, the computed URL is identical to the old release. The filter is purely opt-in for integrators.

**Note on variable naming:** `$root_url = is_ssl() ? 'https' : 'http'` in the old code has been renamed to `$scheme`. This is not a logic change — that value is the scheme argument for the following `set_url_scheme()` call; the variable name was tidied to clarify intent.

### Test coverage

The `get_site_root_url()` unit test covers:

- all four computation branches: single site, multisite main site, multisite subsite via `get_network()`, and the fallback when no network is resolvable;
- the filter contract — that it fires, and receives the computed URL and the current blog ID;
- trailing-slash normalisation of a filtered value returned without one;
- the fallback for every unusable return type (array, object, integer, `null`, `false`, `0`, empty and whitespace-only strings);
- that the filter is applied only once per blog (the caching contract);
- **the blog-keyed cache regression** — after the current blog changes, the URL is recomputed rather than inheriting the previous blog's value. This is the bug the old function-static caused.

`Imagify_Filesystem` could not previously be loaded in the unit suite: it pulls `WP_Filesystem_Base`/`WP_Filesystem_Direct` from `ABSPATH` at file scope, and `yoast/wp-test-utils` defines `ABSPATH` as `true`, turning those into unresolvable relative paths. Redefining `ABSPATH` is not viable — the resulting "Constant already defined" warning leaks into every test running in a separate process and breaks ~59 of them. Instead, minimal stubs for the two WP classes are mirrored into an `ABSPATH`-derived directory that is placed on `include_path` at bootstrap. This unblocks unit-testing other legacy `class-*.php` classes with the same file-scope requires.

### Documentation

- **`docs/imagify-technical-deep-dive.md` (section 14):** Added filter to the hooks reference table with description.
- **`docs/imagify-summary.md` (section 13):** Added filter to the summary hooks table.
- **`readme.txt`:** Added changelog entry under `= 2.3.1 =`.

### New dependencies

None.

### Risks

**Performance:** The filter now fires on every call to `get_site_root_url()` across all code paths. The result is cached per blog, so a callback runs once per blog per process. No measurable impact on sites that don't hook the filter.

**Compatibility:** The default value and behaviour are byte-identical to the current release when no filter is hooked. No breaking changes.

**Unverified domain mapping:** The user-visible fix has not been exercised on a real domain-mapped multisite. See "How to test" — this remains the main pre-merge gate.

**Multisite subsite runtime coverage:** The subsite branch is covered by unit tests with mocked WordPress functions, not by a real multisite bootstrap. Manual verification is still required.

# Mandatory Checklist

## Code validation

  - Validated by unit tests for the filter contract and all computation branches. The domain-mapping scenario itself remains unverified on real infrastructure — see Risks.
  - Every changed line in `get_site_root_url()` is executed by the unit suite.
  - 16 unit tests / 34 assertions. See "Test coverage".

## Code style

  - Filter return values are type- and content-checked, with a fallback to `home_url( '/' )`.
  - Single exit point; inverted branch for clarity.

## Remaining pre-merge gate

The domain-mapping fix has not been manually verified on a real domain-mapped multisite. Everything else — the filter contract, the caching behaviour, the blog-keyed cache fix, and the input validation — is covered by executed automated tests.

# Additional Checks
  - Caching logic is straightforward; comments added at the filter fire point and at the validation step.
  - The filter itself is the observation point: integrators can hook it to inspect the computed root URL per blog.
  - Filter return validated; fallback to safe default.


Tests/Integration/inc/classes/ImagifyFilesystem/TestCase.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace Imagify\Tests\Integration\inc\classes\ImagifyFilesystem;
+
+use Imagify\Tests\Integration\TestCase as BaseTestCase;
+use Imagify_Filesystem;
+
+/**
+ * Base test case for Imagify_Filesystem integration tests.
+ */
```
Miraeld, line 10: **Medium** — `protected $use_api = false;` does not override the base `Tests\Integration\TestCase::$useApi` (different casing/underscore). The base class still sees its own `$useApi = true` default, so its `set_up()`/`tear_down()` API-related singleton reset (`setSingletonInstance(Imagify::class, null)`) and option juggling still run for every test in this suite — the intent of skipping that setup for a filesystem-only test case is silently defeated.\n\n**Fix:** rename the property to `$useApi` to actually match and override the parent declaration.

Honemo (approved):

PR #1202, opened 2026-07-27
Closes #1011: fix next-gen generation tool never completing when the converted file is heavier

### LINKED ISSUE #1011: Next-Gen Generation Tool Stuck When AVIF or WebP Is Heavier Than Optimized Original Image
Context
Along with some other issues that cause glitches in the Next-Gen generation tool in the Imagify plugin settings, this is one of them.

When a small-size image is uploaded and Imagify tries to generate next-gen formats (WebP or AVIF), there are cases where the generated WebP or AVIF file is heavier than the optimized standard format image.

In such cases, Imagify does not generate the next-gen version and shows a message like:
	•	“AVIF is less performant than the original image”
	•	Similar message for WebP

When this happens, the Imagify settings page shows “Generate missing next-gen versions”, but the process never completes. It stays stuck indefinitely because Imagify cannot generate next-gen versions for those images.

This creates confusion for users and leads to support tickets, especially on sites with a large number of images.

Steps to Reproduce (STR)
	1.	Enable Lossless compression in Imagify settings.
	2.	Enable AVIF format.
	3.	Upload the attached example images.
	4.	Let the images auto-optimize.
	5.	Check the Media Library. You will see the message saying AVIF is less performant than the original image.

Image

	6.	Go to Imagify settings and click “Generate missing next-gen versions”.

Image

	7.	The process remains stuck indefinitely.

Example images:

Image

Image

Note:
The same behavior also happens with WebP when WebP is heavier than the optimized original image. I do not have a WebP example image at the moment, but the behavior is identical.

Image

Additional Related Case
There is a similar but slightly different case:

Some images are already optimized, and Imagify correctly shows the message:
“Well done. This image is already optimized.”

However, if AVIF (or WebP) would be heavier for such images, the “already optimized” message is shown, but no next-gen file is generated in the background. These images are still counted as “missing next-gen versions”, which again causes the Next-Gen generation tool to stay stuck.

Image

An example image is attached to test this scenario as well. The steps to reproduce are similar.
Example image: 

Image

Expected Behavior
One of the following should happen to avoid confusion and the stuck state:
	•	Clearly exclude such images from the “missing next-gen versions” count, or
	•	Allow generating next-gen images even if they are slightly heavier (this is also a feature request from some users), or
	•	Show a clear and explicit message explaining why next-gen versions cannot be generated and that these images are excluded from the process.

Actual Behavior
	•	Next-gen versions are not generated.
	•	Images remain counted as missing next-gen versions.
	•	The “Generate missing next-gen versions” tool never completes.
	•	Users are confused and think the tool is broken.

Impact
	•	Infinite or stalled Next-Gen generation process.
	•	Confusing UX with no clear explanation.
	•	Increased support tickets, especially on sites with many images.

Although we chose Lossless compression and smaller image sizes for testing, the issue could occur with regular images and smart compression as well if the Imagify algorithm is unable to optimize further or the generated next-gen is heavier.

Miraeld:
# Description

Fixes #1011

"Generate missing next-gen versions" never finishes when a WebP or AVIF version would be heavier than the optimized image. Affected media stay in the "missing next-gen versions" count forever, the progress bar never reaches the end, and every run re-uploads those files to the API — spending credits on conversions that cannot succeed. Users read this as a broken tool and open support tickets.

After this PR the tool completes, those media are no longer counted as missing, and the media library shows the reason the API gave instead of silently skipping them.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

**Automated — unit (`composer test-unit`), all passing:**

Baseline on `develop` was 345 tests / 798 assertions / 17 risky. With this branch: **350 tests / 814 assertions / 17 risky, 0 failures.** The delta is exactly the 5 new tests; the 17 risky are pre-existing `ReflectionProperty::setValue()` deprecation warnings, identical on both branches.

New file `Tests/Unit/classes/Optimization/Process/AbstractProcess/update_size_optimization_data.php`:

| Test | What it pins |
|---|---|
| `testStoresPermanentErrorUnderSuffixedKeyForNextGenSize` | Refused conversion is stored under `full@imagify-avif`, `success => false`, flagged |
| `testDoesNotOverwriteBaseSizeWhenNextGenIsRefused` | The base size's optimization record is left alone |
| `testKeepsPreviousBehaviourForNonNextGenSize` | Plain sizes carrying a message behave exactly as before |
| `testKeepsTransientNextGenErrorRetryable` | A `WP_Error` failure is stored **without** the flag, so it is still retried |
| `testStoresSuccessfulNextGenConversionUnchanged` | A normal successful conversion is untouched |

The fourth is the important one — it pins the distinction the queries rely on to decide what may be retried.

**Automated — serialization check.** The stored shape and the SQL predicate were verified against real `serialize()` output:

```
permanent: a:1:{s:17:"full@imagify-avif";a:3:{s:15:"permanent_error";b:1;s:7:"success";b:0;s:5:"error";s:43:"…";}}  → matched by the predicate
transient: a:1:{s:17:"full@imagify-avif";a:2:{s:7:"success";b:0;s:5:"error";s:10:"cURL error";}}                    → not matched, stays retryable
```

**Lint / static analysis.** `vendor/bin/phpcs` on all 9 changed production files → 0 violations. `composer run-stan` → 2 errors, both pre-existing in `classes/Media/WP.php:214` and both present on `develop`; **no new errors**.

**Not run — integration.** `Tests/Integration/classes/Bulk/WP/GetOptimizedMediaIdsWithoutFormatTest.php` covers the query-level exclusion but could not be executed here (no WordPress test installation configured). **It needs a CI run before this is considered verified.**

**Not run — manual reproduction.** The end-to-end scenario from the issue was not reproduced against a live Imagify API. See "Unticked items justification".

### How to test

1. Enable Lossless compression and the AVIF format in Imagify settings.
2. Upload the small example images attached to #1011 and let them auto-optimize.
3. In the Media Library, confirm the message saying the next-gen version is larger than the original, and that a **"Next-Gen status:"** line now shows that reason.
4. Go to Imagify settings and click **"Generate missing next-gen versions"**.
5. **Expected:** the process completes instead of hanging, and those images are no longer counted as missing.
6. Repeat the click and confirm from your Imagify account that no further credits are consumed for those images.
7. **Regression:** on media where next-gen conversion succeeds normally, confirm the files are still generated and the counts still drop to zero as before.
8. **Regression (retry path):** simulate a transient failure (block outgoing HTTP, or exhaust quota) on a convertible image, restore normal conditions, and confirm the image is picked up again on the next run — it must **not** be permanently skipped.

Step 8 is the one most worth QA attention: it is the behaviour that separates this fix from simply hiding failures.

- **"Generate missing next-gen versions" bulk tool** — the reported symptom.
- **Missing next-gen counts everywhere** — a single query feeds the settings AJAX tool, the settings-save hook, the WP-CLI `generate-missing-nextgen` command, the `GenerateMissingNextgen` and `GetNextgenCoverage` abilities, and the Imagifybeat progress poll. All six change behaviour together.
- **Media library optimization details** — a new "Next-Gen status:" line.
- **Custom folders and NextGEN Gallery** — each has its own data writer and query; both were updated in step with the media library context.
- **Optimization data written per size** — the base size record is no longer overwritten by a refused next-gen attempt, so displayed per-size figures change for affected media.

### Documentation

The API answers a refused conversion with **HTTP 200 and a `message`**, not an error. `AbstractProcess::update_size_optimization_data()` therefore took its success branch, and then stripped the next-gen suffix from the size name:

```php
if ( property_exists( $response, 'message' ) ) {
	$size = str_replace( $this->format, '', $size );
}
```

So `full@imagify-avif` was written as `full`. No next-gen entry ever existed, and the base size's genuine optimization record was overwritten by the refused attempt's figures.

The bulk queries exclude a media only when its data contains a *successful* next-gen entry (`<suffix>";a:4:{s:7:"success";b:1;`). With no entry at all, affected media were re-selected on every run. Imagifybeat reports that same query re-executed as "remaining", and the JS stops only at `remaining === 0` — so the bar could never complete, and each pass re-uploaded the files.

The fix keeps the suffix and records a terminal entry under the next-gen size name:

```php
if ( $this->is_next_gen_size( $size ) ) {
	$data['success']         = false;
	$data['error']           = $data['message'];
	$data['permanent_error'] = true;
} else {
	$size = str_replace( $this->format, '', $size );
}
```

Non-next-gen sizes are untouched — a plain `full` size legitimately carries "already compressed" messages today.

**Why a flag rather than just `success => false`.** Transient failures (network, timeout, quota) already write `success => false` under the suffixed key and are correctly retried. Marking permanent refusals identically would strand an image forever after a single API blip. The flag is written as the **first** key of the size entry so the serialized value has a deterministic prefix; with `success`/`error` first, the variable-length error string sits in the middle and cannot be matched. The queries then add:

```php
'%' . $wpdb->esc_like( $nextgen_suffix . '";a:3:{s:15:"permanent_error";b:1;' ) . '%'
```

Matching serialized data with `LIKE` is fragile, and the code says so in a comment at both the write and read sites: the key order must not change. This follows the pattern already used by the existing predicate rather than introducing a new meta key, which would have required a migration on every install plus parallel handling in the custom-folders table.

**No backfill migration**, deliberately. The database never recorded *why* a next-gen version was missing, so a migration would have to guess, and guessing wrong permanently strands images that would convert successfully. The marker is written on the next attempt instead: self-healing, and it bounds the wasted credits at one further pass per image.

### New dependencies

None.

### Risks

**Retry behaviour is the main risk.** If the permanent/transient distinction were wrong, images would be skipped forever. This is covered by `testKeepsTransientNextGenErrorRetryable` and by the serialization check above, and is called out as step 8 of "How to test".

**Serialized `LIKE` matching is fragile.** Changing the key order in the data writers silently breaks the exclusion. Commented at all four sites. The pre-existing predicate has the same property.

**Per-size figures change for affected media.** Because the base size record is no longer overwritten, media previously showing the refused attempt's numbers will show their genuine optimization data. This is a correction, but it is a visible change.

**Performance:** one additional `NOT LIKE` on an already-`LIKE`-filtered query. No new joins, no new meta reads. Net positive in practice, since affected media stop being re-uploaded on every run.

**Security:** none. No new input, output, or capability surface. The added predicate is built with `esc_like()` and `prepare()` like the existing one; the displayed reason is escaped with `esc_html()`.

# Mandatory Checklist

## Code validation

- [ ] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.

## Code style

**"I validated all the Acceptance Criteria"** is unticked because the end-to-end scenario from the issue was not reproduced against a live Imagify API, and the integration test covering the query-level exclusion could not be executed (no WordPress test installation in this environment). The unit tests and the serialization check establish the logic; they do not establish the live behaviour. Both gaps are QA gates, listed under "How to test".

Two further items need confirmation and are flagged rather than assumed:

1. **The AVIF message wording is unverified.** Only the WebP string was mapped, so AVIF users saw the raw untranslated API string. Two casings are mapped following the WebP pattern, with a comment recording that the exact wording was not confirmed against a live response. If the API returns different wording, the message falls through untranslated — the same behaviour as today, so this is not a regression, but it should be confirmed.
2. **`$data['status']` is left as `'success'`** on the refusal path even though `success` is now `false`. Changing it risks affecting the top-level `_imagify_status` meta and the existing "Convert:" message display, which is beyond the scope of this fix. Worth a reviewer's opinion.

# Additional Checks
  - The stored `permanent_error` flag and its reason are readable in `_imagify_data` and surfaced in the media library, so an affected media can be diagnosed without reproducing the API call.
  - This PR is itself the error-handling path for a refused API conversion.

### DIFF
--- a/Tests/Integration/classes/Bulk/WP/GetOptimizedMediaIdsWithoutFormatTest.php
+++ b/Tests/Integration/classes/Bulk/WP/GetOptimizedMediaIdsWithoutFormatTest.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace Imagify\Tests\Integration\classes\Bulk\WP;
+
+use Imagify\Bulk\WP as BulkWP;
+use Imagify\Tests\Integration\TestCase;
+
+/**
+ * Tests for \Imagify\Bulk\WP::get_optimized_media_ids_without_format().
+ *
+ * @covers \Imagify\Bulk\WP::get_optimized_media_ids_without_format
+ * @group  NextGenPermanentError
+ */
+class GetOptimizedMediaIdsWithoutFormatTest extends TestCase {
+	protected $useApi = false;
+
+	/**
+	 * Creates an optimized attachment carrying the given `_imagify_data` sizes.
+	 *
+	 * @param array $sizes The `sizes` sub-array of `_imagify_data`.
+	 *
+	 * @return int
+	 */
+	private function create_optimized_attachment( array $sizes ): int {
+		$attachment_id = $this->factory()->attachment->create_object(
+			[
+				'file'           => 'image.jpg',
+				'post_mime_type' => 'image/jpeg',
+				'post_status'    => 'inherit',
+			]
+		);
+
+		update_post_meta( $attachment_id, '_wp_attached_file', 'image.jpg' );
+		update_post_meta( $attachment_id, '_wp_attachment_metadata', [ 'file' => 'image.jpg' ] );
+		update_post_meta( $attachment_id, '_imagify_status', 'success' );
+		update_post_meta(
+			$attachment_id,
+			'_imagify_data',
+			[
+				'sizes' => $sizes,
+			]
+		);
+
+		return $attachment_id;
+	}
+
+	/**
+	 * Test: a media whose AVIF conversion was permanently refused is not returned any more.
+	 */
+	public function testExcludesMediaWithPermanentNextGenError() {
+		$attachment_id = $this->create_optimized_attachment(
+			[
+				'full'               => [
+					'success'        => true,
+					'original_size'  => 1000,
+					'optimized_size' => 800,
+					'percent'        => 20.0,
+				],
+				'full@imagify-avif' => [
+					'permanent_error' => true,
+					'success'         => false,
+					'error'           => 'AVIF file is larger than the original image',
+				],
+			]
+		);
+
+		$result = ( new BulkWP() )->get_optimized_media_ids_without_format( 'avif' );
+
+		$this->assertNotContains( $attachment_id, $result['ids'] );
+	}
+
+	/**
+	 * Test: a media whose AVIF conversion failed transiently is still returned, so it gets retried.
+	 */
+	public function testKeepsMediaWithTransientNextGenError() {
+		$attachment_id = $this->create_optimized_attachment(
+			[
+				'full'               => [
+					'success'        => true,
+					'original_size'  => 1000,
+					'optimized_size' => 800,
+					'percent'        => 20.0,
+				],
+				'full@imagify-avif' => [
+					'success' => false,
+					'error'   => 'cURL error 28: Operation timed out',
+				],
+			]
+		);
+
+		$result = ( new BulkWP() )->get_optimized_media_ids_without_format( 'avif' );
+
+		$this->assertContains( $attachment_id, $result['ids'] );
+	}
+}
--- a/Tests/Unit/classes/Optimization/Process/AbstractProcess/update_size_optimization_data.php
+++ b/Tests/Unit/classes/Optimization/Process/AbstractProcess/update_size_optimization_data.php
@@ -1,0 +1,199 @@
+<?php
+declare(strict_types=1);
+
+namespace Imagify\Tests\Unit\classes\Optimization\Process\AbstractProcess;
+
+use Brain\Monkey\Functions;
+use Imagify\Optimization\Data\DataInterface;
+use Imagify\Optimization\Process\AbstractProcess;
+use Imagify\Tests\Unit\TestCase;
+use Mockery;
+use WP_Error;
+
+/**
+ * Tests for \Imagify\Optimization\Process\AbstractProcess::update_size_optimization_data().
+ *
+ * @covers \Imagify\Optimization\Process\AbstractProcess::update_size_optimization_data
+ * @group  NextGenPermanentError
+ */
+class Test_UpdateSizeOptimizationData extends TestCase {
+
+	/**
+	 * Stores the arguments the data layer received.
+	 *
+	 * @var array
+	 */
+	private $stored = [];
+
+	/**
+	 * Sets up the common function stubs.
+	 *
+	 * @return void
+	 */
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->stored = [];
+
+		Functions\when( 'imagify_translate_api_message' )->returnArg();
+	}
+
+	/**
+	 * Builds a process whose data layer records what it is asked to store.
+	 *
+	 * @return ProcessStub
+	 */
+	private function get_process(): ProcessStub {
+		$data = Mockery::mock( DataInterface::class );
+
+		$data->shouldReceive( 'update_size_optimization_data' )
+			->andReturnUsing(
+				function ( $size, $size_data ) {
+					$this->stored[ $size ] = $size_data;
+				}
+			);
+
+		return new ProcessStub( $data, ProcessStub::AVIF_SUFFIX );
+	}
+
+	/**
+	 * Test: a next-gen size refused by the API is stored under the suffixed key as a permanent error.
+	 */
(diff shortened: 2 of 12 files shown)


Miraeld:
Closing this: I opened it based on an old version of `develop` by mistake, so the whole premise is wrong.

My local `develop` was 7 commits behind `origin/develop`, and I did not fetch before branching. The fix here targets a suffix-stripping block in `AbstractProcess::update_size_optimization_data()` that #1194 already removed three days ago:

```php
-		if ( property_exists( $response, 'message' ) ) {
-			$size = str_replace( $this->format, '', $size );
-		}
```

On current `develop` the suffix is kept, so a refused conversion is stored under `full@imagify-avif` and serializes as `a:4:{s:7:"success";b:1;` — exactly what the bulk query's `NOT LIKE` excludes. The stuck loop described in #1011 is therefore already fixed, and this PR would only re-solve it.

Not rebasing, because the fix is redundant rather than out of date. There is a separate, real issue with how #1194 leaves that data (it records a successful next-gen version for a file that was never written) — that will get its own PR rather than riding on this branch.

Sorry for the noise.

PR #1203, opened 2026-07-27
Refs #1011: stop reporting a next-gen version when none was generated

### LINKED ISSUE #1011: Next-Gen Generation Tool Stuck When AVIF or WebP Is Heavier Than Optimized Original Image

Miraeld:
# Description

Refs #1011

Since #1194, Imagify reports a next-gen version as generated for images where no next-gen file exists on disk.

When the API declines a conversion — because the WebP/AVIF version would be heavier than the optimized image, or the image is already compressed — it answers with **HTTP 200 and a `message`** rather than an error. `Imagify\Optimization\File::optimize()` correctly writes no next-gen file in that case, but the optimization data records the size as `success => true`. `has_next_gen()` reads that same record, so the media library shows **"Next-Gen generated"** for a file that was never created, and the size's stats are populated from a conversion that produced nothing.

This PR records those refusals as what they are — a permanent refusal, with the reason the API gave — while keeping the media out of the "missing next-gen versions" count so the bulk tool still completes.

**This does not undo #1194.** That fix made the bulk tool terminate by having the refused size match the query's "already has next-gen" pattern. Here the media is excluded explicitly instead, by a second predicate matching the refusal marker. The tool still completes; it just no longer relies on the data claiming a success that did not happen.

## Type of change

- [ ] Sub-task of #1011
- [ ] Chore
- [ ] Release

### What was tested

**Manual — reproduced end to end against a live Imagify API.** Local WP install, PHP 8.4, real API key, `lossless = 1`, `optimization_format = avif`, backups enabled. Steps as run:

1. Imported a small PNG (attachment **115**) and let it auto-optimize.
2. Inspected `_imagify_data`. The three `…@imagify-avif` sizes recorded:
   `success = false`, `permanent_error = true`, `error = "AVIF file is larger than the original image"`.
   The non-next-gen sizes (`medium`, `thumbnail`) recorded `"WELL DONE. This image is already compressed…"` **without** `permanent_error`, confirming the marker is applied only to next-gen sizes.
3. `has_next_gen()` returned **false**, and no `.avif` file existed on disk. The false claim is gone.
4. Rendered `get_imagify_attachment_optimization_text()` for that media. Output contains:
   `Next-Gen generated: No` … `Next-Gen status: AVIF file is larger than the original image`.
5. `get_optimized_media_ids_without_format( 'avif' )` → attachment 115 **excluded** from the missing-next-gen selection. The count no longer grows forever.
6. **Regression, normal conversions (step 7 below):** imported a 3456×2234 PNG (attachment **116**) → AVIF sizes succeeded, `permanent_error = 0`, `has_next_gen()` **true**, and the `.avif` files were written to disk.
7. **Regression, retry path (step 8 below).** Injected a `WP_Error` on next-gen sizes via `imagify_before_optimize_size` to simulate a transient failure, then imported another large PNG (attachment **118**):
   - sizes recorded `success = false`, **`permanent_error` unset** → correctly treated as transient
   - `get_optimized_media_ids_without_format( 'avif' )` **still selected** attachment 118 → retryable, not stranded
8. Removed the simulated failure and called `generate_nextgen_versions()` on 118 → **7/7 AVIF sizes converted and 7 `.avif` files written.** The retry path completes.

**Result: the permanent/transient discrimination — the core risk in this PR — behaves exactly as designed.** Permanent refusals are excluded from the count; transient failures remain retryable and convert successfully once conditions recover.

**The guessed AVIF wording is now confirmed.** This PR adds `'Avif is less performant than original'` with a comment noting the wording was unverified against a live response. It is verified: that is the exact raw string the API returns, and the translated message *"AVIF file is larger than the original image"* rendered correctly in step 2, which only happens if the mapping matched. The uppercase `'AVIF is less performant than original'` fallback appears unnecessary, though harmless.

*Limitation, stated rather than glossed:* the transient failure in step 7 was injected at the `imagify_before_optimize_size` filter rather than by a real network fault, because Imagify uploads use `Imagify::curl_http_call()` directly and bypass the WP HTTP API, so `pre_http_request` cannot intercept them. Both routes converge on the same `is_wp_error( $response )` branch of `update_size_optimization_data()`, which is the discrimination under test.

### How to test

1. Enable Lossless compression and the AVIF format in Imagify settings.
2. Upload the small example images attached to #1011 and let them auto-optimize.
3. In the Media Library, check the optimization details for one of them.
   - **Before:** "Next-Gen generated: yes", despite no `.avif` file existing next to the image.
   - **After:** no false claim, plus a "Next-Gen status:" line giving the reason returned by the API.
   - Confirm on disk that there is indeed no `.avif` file for that image.
4. Go to Imagify settings and click **"Generate missing next-gen versions"**.
5. **Expected:** the process still completes — this is the #1194 behaviour that must not regress — and those images are not counted as missing.
6. Click it again and confirm from your Imagify account that no further credits are spent on those images.
7. **Regression — normal conversions:** on images where next-gen conversion succeeds, confirm the files are still written, still reported as generated, and the counts still reach zero.
8. **Regression — retry path:** block outgoing HTTP (or exhaust quota) so a convertible image fails transiently, restore normal conditions, then re-run the tool. The image **must** be picked up again and converted. It must not be permanently skipped.

Steps 5 and 8 are the two worth the most QA attention: 5 guards against regressing #1194, 8 guards against the fix silently stranding retryable images.

- **Media library optimization details** — no longer claims a next-gen version that does not exist; adds a "Next-Gen status:" line with the reason.
- **"Generate missing next-gen versions" tool** — still completes, now via an explicit exclusion rather than a false success record.
- **Missing next-gen counts everywhere** — one query feeds the settings AJAX tool, the settings-save hook, the WP-CLI command, the `GenerateMissingNextgen` and `GetNextgenCoverage` abilities, and the Imagifybeat progress poll. All change together.
- **Custom folders and NextGEN Gallery** — each has its own data writer and query; both updated in step.
- **Per-size optimization figures** — refused next-gen sizes no longer carry `original_size` / `optimized_size` from a conversion that produced no file.

### Documentation

In `AbstractProcess::update_size_optimization_data()`, a 200-with-`message` response on a next-gen size is now recorded as a terminal refusal rather than a success:

```php
if ( property_exists( $response, 'message' ) && $this->is_next_gen_size( $size ) ) {
	$data['success']         = false;
	$data['error']           = $data['message'];
	$data['permanent_error'] = true;
}
```

Non-next-gen sizes are untouched: a plain `full` size legitimately carries "already compressed" messages.

**Why a flag rather than plain `success => false`.** Transient failures (network, timeout, quota) already write `success => false` under the suffixed key and are correctly retried. Marking refusals identically would strand an image forever after a single API blip, and would also re-break #1194 by putting the media back into the missing count. The flag is written as the **first** key of the size entry so the serialized value has a deterministic prefix; with `success`/`error` first, the variable-length error string sits in the middle and cannot be matched. The queries then add:

```php
'%' . $wpdb->esc_like( $nextgen_suffix . '";a:3:{s:15:"permanent_error";b:1;' ) . '%'
```

Matching serialized data with `LIKE` is fragile, and both the write and read sites carry a comment saying the key order must not change. This follows the pattern of the existing predicate rather than introducing a new meta key, which would need a migration on every install plus parallel handling in the custom-folders table.

**No backfill migration**, deliberately. Media already in the false-success state will be re-evaluated on their next pass and recorded correctly, at a cost of one further API call each. A migration would have to guess which existing records came from a refusal, and guessing wrong would strand images that would convert successfully.

### New dependencies

None.

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

N/A

# Additional Checks
  - The `permanent_error` flag and its reason are readable in `_imagify_data` and surfaced in the media library, so an affected media can be diagnosed without reproducing the API call.

### DIFF
--- a/Tests/Integration/classes/Bulk/WP/GetOptimizedMediaIdsWithoutFormatTest.php
+++ b/Tests/Integration/classes/Bulk/WP/GetOptimizedMediaIdsWithoutFormatTest.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace Imagify\Tests\Integration\classes\Bulk\WP;
+
+use Imagify\Bulk\WP as BulkWP;
+use Imagify\Tests\Integration\TestCase;
+
+/**
+ * Tests for \Imagify\Bulk\WP::get_optimized_media_ids_without_format().
+ *
+ * @covers \Imagify\Bulk\WP::get_optimized_media_ids_without_format
+ * @group  NextGenPermanentError
+ */
+class GetOptimizedMediaIdsWithoutFormatTest extends TestCase {
+	protected $useApi = false;
+
+	/**
+	 * Creates an optimized attachment carrying the given `_imagify_data` sizes.
+	 *
+	 * @param array $sizes The `sizes` sub-array of `_imagify_data`.
+	 *
+	 * @return int
+	 */
+	private function create_optimized_attachment( array $sizes ): int {
+		$attachment_id = $this->factory()->attachment->create_object(
+			[
+				'file'           => 'image.jpg',
+				'post_mime_type' => 'image/jpeg',
+				'post_status'    => 'inherit',
+			]
+		);
+
+		update_post_meta( $attachment_id, '_wp_attached_file', 'image.jpg' );
+		update_post_meta( $attachment_id, '_wp_attachment_metadata', [ 'file' => 'image.jpg' ] );
+		update_post_meta( $attachment_id, '_imagify_status', 'success' );
+		update_post_meta(
+			$attachment_id,
+			'_imagify_data',
+			[
+				'sizes' => $sizes,
+			]
+		);
+
+		return $attachment_id;
+	}
+
+	/**
+	 * Test: a media whose AVIF conversion was permanently refused is not returned any more.
+	 */
+	public function testExcludesMediaWithPermanentNextGenError() {
+		$attachment_id = $this->create_optimized_attachment(
+			[
+				'full'               => [
+					'success'        => true,
+					'original_size'  => 1000,
+					'optimized_size' => 800,
+					'percent'        => 20.0,
+				],
+				'full@imagify-avif' => [
+					'permanent_error' => true,
+					'success'         => false,
+					'error'           => 'AVIF file is larger than the original image',
+				],
+			]
+		);
+
+		$result = ( new BulkWP() )->get_optimized_media_ids_without_format( 'avif' );
+
+		$this->assertNotContains( $attachment_id, $result['ids'] );
+	}
+
+	/**
+	 * Test: a media whose AVIF conversion failed transiently is still returned, so it gets retried.
+	 */
+	public function testKeepsMediaWithTransientNextGenError() {
+		$attachment_id = $this->create_optimized_attachment(
+			[
+				'full'               => [
+					'success'        => true,
+					'original_size'  => 1000,
+					'optimized_size' => 800,
+					'percent'        => 20.0,
+				],
+				'full@imagify-avif' => [
+					'success' => false,
+					'error'   => 'cURL error 28: Operation timed out',
+				],
+			]
+		);
+
+		$result = ( new BulkWP() )->get_optimized_media_ids_without_format( 'avif' );
+
+		$this->assertContains( $attachment_id, $result['ids'] );
+	}
+}
--- a/Tests/Integration/classes/Bulk/WP/GetOptimizedMediaIdsWithoutFormatTest.php
+++ b/Tests/Integration/classes/Bulk/WP/GetOptimizedMediaIdsWithoutFormatTest.php
@@ -12,26 +12,69 @@ use Imagify\Tests\Integration\TestCase;
  * @group  NextGenPermanentError
  */
 class GetOptimizedMediaIdsWithoutFormatTest extends TestCase {
-	protected $useApi = false;
+	protected $useApi = false; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
+
+	/**
+	 * Absolute paths created by the fixtures, removed in tear_down().
+	 *
+	 * @var array
+	 */
+	protected $created_files = [];
+
+	/**
+	 * Cleans up the files created by the fixtures.
+	 */
+	public function tear_down() {
+		foreach ( $this->created_files as $path ) {
+			if ( file_exists( $path ) ) {
+				unlink( $path );
+			}
+		}
+
+		$this->created_files = [];
+
+		parent::tear_down();
+	}
 
 	/**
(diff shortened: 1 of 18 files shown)


Honemo (approved):

Miraeld:
As it's test by dev and manual testing confirmed it works, I'm moving it to `QA Done`

Issue #641, opened 2021-12-10
Sometimes Optimization Can Hang For a Long Time and Then Fail


joejoe04:
**Describe the bug**
Sometimes, when optimization is attempted on an image, it can just hang on "Optimizing..." for a long time:

![image](https://user-images.githubusercontent.com/45705494/145593895-b33da66b-fb02-4357-bd38-a9c5b8467cba.png)

And then optimization fails, but instead of offering an error message to give the user an indication of what is wrong, it just fails and the Optimize button returns:

![image](https://user-images.githubusercontent.com/45705494/145594191-58a9de76-2168-4911-b25f-301154b94fa4.png)

We can see here that the process is set to run up to 10 minutes:
https://github.com/wp-media/imagify-plugin/blob/5971929e9842066e1351704ce14f25c5f31a7180/classes/Optimization/Process/AbstractProcess.php#L1710

**To Reproduce**
Unfortunately, not sure how to reproduce this.

**Expected behavior**
We should see if there is a way to prevent optimization from hanging if possible. At the least, an error message should be returned to let the user know what is happening and possible solutions.

Could also help if the functionality suggested in the following issue is implemented so optimization can be stopped manually by the user: https://github.com/wp-media/imagify-plugin/issues/642

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

PR #1208, opened 2026-07-29
Add a manual stop for the bulk optimization

### LINKED ISSUE #642: Add an Option to Manually Stop Optimization of Images
When optimization is not working for various reasons, the user can see the "Optimizing..." happening and it can take a long time for it to fail:

image

This could be due to issues like cURL error 28 / timeouts. It could be related to a possibly different issue where optimization hangs for many minutes without returning any error: https://github.com/wp-media/imagify-plugin/issues/641

If either the user or support team is trying to debug an issue with optimization, when you try optimizing an image, it can be clear that whatever changes were tried are not working after so long, but there's currently no way to stop the optimization attempt manually.

It could be helpful in some cases to have the ability to stop or cancel the optimization attempt manually.

Miraeld:
# Description

Fixes #642

When a bulk optimization goes wrong — cURL timeouts, an unresponsive API, the hang described in #641 — the user is stuck watching "Optimizing…" with no way out. Today the only escape is deactivating the plugin (as reported in the issue comment), which leaves the Action Scheduler queue and the progress transients behind.

This PR adds a **Stop the optimization** button on the bulk optimization page. It cancels every media still waiting in the queue, clears the progress state, and hands the page back to the user in a clean, restartable state.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated (unit, `--group BulkRunStop`, 9 tests green). Coverage follows the repository's fixture pattern: `Tests/Unit/classes/Bulk/runStop.php` with its data cases in `Tests/Fixtures/classes/Bulk/runStop.php`, consumed through `configTestData`.

Data-provider cases (the `run_stop()` return contract):

- Nothing running → `not-running`, `cancelled` 0.
- One running context → the remaining media are reported as `cancelled`.
- Two running contexts → the remaining counts are summed.
- One running, one idle context → only the running one contributes.
- A corrupted (negative) `remaining` counter never yields a negative `cancelled`.

Behavioral cases:

- Both bulk hooks are flushed for every context, with the exact expected hook/group pairs (`imagify_optimize_media|imagify-wp-optimize-media`, `imagify_convert_next_gen|imagify-wp-convert-nextgen`, and the `custom-folders` equivalents).
- `imagify_wp_optimize_running`, `imagify_bulk_optimization_result` and `imagify_missing_next_gen_total` are all deleted.
- `imagify_bulk_stopped` fires when a run was stopped, and does **not** fire when nothing was running.

Full unit suite: **368 tests, 846 assertions, no failures** (up from 359/837 on `develop`). `phpcs` clean on every changed PHP file. The 2 remaining PHPStan errors are pre-existing in `classes/Media/WP.php` and unrelated. `assets/admin/js/bulk.js` was rebuilt with `bud build` from `_dev/src/bulk.js`.

Manual: the stop flow was exercised on a local site and behaves as described — the button appears during a run, the confirmation reports the cancelled media, and the page returns to a restartable state.

### How to test

1. Check out this branch on a site with a connected Imagify API key and a few hundred unoptimized media (`bash bin/dev-seed.sh` helps).
2. Go to **Media → Bulk Optimization**, select the library, click **Imagif'em all**.
3. Once the progress bar moves, a **Stop the optimization** button appears next to the (now disabled) main button. Click it.
4. Confirm in the dialog. Expected: a confirmation reporting how many media were removed from the queue, the progress bar collapses, and **Imagif'em all** becomes clickable again.
5. Verify the queue is actually empty — `wp action-scheduler list --group=imagify-wp-optimize-media --status=pending` returns nothing (or check **Tools → Scheduled Actions**).
6. Verify the counters are clean — `wp transient get imagify_wp_optimize_running` returns nothing.
7. Restart the optimization. It must start from a fresh progress bar (0%) and pick up the media that were skipped.
8. Reload the bulk page **while a run is in progress**: the Stop button must already be visible on load (it is rendered server-side from the running transients).
9. Simulate the issue's scenario: block outgoing HTTP to Imagify (e.g. `define( 'WP_HTTP_BLOCK_EXTERNAL', true );`), start a bulk run, and confirm the run can be stopped without waiting for the timeouts.
10. Check permissions: a user without the `bulk-optimize` capability calling `admin-ajax.php?action=imagify_bulk_stop` must be rejected.

- **Bulk optimization page** (`Media → Bulk Optimization`) — the whole start/progress/complete cycle. The Stop button is additive, but it shares `queueEmpty` with the normal end-of-run path, so the regular "run to completion" flow needs a smoke check: progress bar, share/summary box, stats refresh, and the re-enabling of the main button.
- **Action Scheduler queue** — only the `imagify-{context}-optimize-media` and `imagify-{context}-convert-nextgen` groups are touched, and only their **pending** actions. Nothing else in the queue is affected.
- **Bulk progress transients** — `imagify_{context}_optimize_running`, `imagify_bulk_optimization_result`, `imagify_missing_next_gen_total`.
- **Bulk complete admin notice** — a manual stop deliberately does *not* set `imagify_bulk_optimization_complete`, so no "optimization complete" notice is shown after a stop. Worth confirming the notice still appears after a *normal* full run.
- Untouched: single-image optimization from the media library, restore, settings, WP-CLI, MCP abilities.

### Documentation

No developer doc page covers the bulk AJAX endpoints today, so nothing under `docs/` needed an update. The new `imagify_bulk_stopped` action is documented inline with a full docblock, consistent with the other bulk hooks.

**What was added**

`Imagify\Bulk\Bulk::run_stop( array $contexts )` — for each context it reads the `imagify_{context}_optimize_running` transient to learn how many media are still queued, unschedules the pending actions of both bulk hooks, and deletes the progress transients. It returns `[ 'success', 'message', 'cancelled' ]`, mirroring the shape already used by `run_optimize()` and `run_restore()`.

`Bulk::bulk_stop_callback()` — the `wp_ajax_imagify_bulk_stop` handler. Nonce check via `imagify_check_nonce( 'imagify-bulk-optimize' )` (the nonce the bulk page already ships) plus a `bulk-optimize` capability check on **every** context before doing anything, same as `missing_nextgen_callback()`.

`Bulk::cancel_pending_actions()` — a small private wrapper over `as_unschedule_all_actions( $hook, [], $group )`, guarded by `function_exists()` and a `try/catch`, following the existing `as_enqueue_async_action()` call sites in this class.

**Why the count comes from the transient**

`as_unschedule_all_actions()` returns nothing, and counting the cancelled actions would mean either an extra unbounded `as_get_scheduled_actions()` query on a potentially huge queue, or reaching into `ActionScheduler::store()` internals. The `remaining` value in the progress transient is the exact same number the progress bar is already driven by, so it is both cheaper and consistent with what the user sees.

**Why `imagify_bulk_optimization_result` is deleted**

That transient is the running "N media optimized, X bytes saved" tally, and it is normally cleared when the *bulk complete* notice renders. Since a manual stop does not raise that notice, leaving the tally behind would leak the stopped run's numbers into the next run's summary. Deleting it keeps a restarted run honest.

**Frontend** — `_dev/src/bulk.js` gains `toggleStopButton()`, `maybeStopAllProcesses()` (SweetAlert confirmation, reusing the existing dialog conventions), `stopAllProcesses()` (the AJAX call) and `endStoppedProcess()` (sets `processIsStopped`, empties the local queue, triggers `queueEmpty.imagify` so the page resolves through the exact same teardown path as a completed run). The button is shown when a run starts and hidden by `queueEmpty`. `views/page-bulk.php` renders it hidden unless a run is already in progress, so reloading mid-run still offers the stop.

### New dependencies

None.

### Risks

- **Nothing can interrupt an in-flight optimization.** Action Scheduler cannot kill a running action, and the Imagify API call is a blocking HTTP request. The media being processed at the moment of the stop still completes (and still consumes a credit). This is stated explicitly in the confirmation dialog rather than hidden.
- **Cancelled actions are cancelled, not deleted.** They stay visible in Action Scheduler with the `canceled` status, which is the normal Action Scheduler lifecycle and keeps the run auditable.
- **Concurrent runs across contexts.** `get_contexts()` returns every context available in the current admin, so the stop applies to library *and* custom folders at once. That matches the bulk page, where a single button starts them together.
- **Multisite / network admin.** `get_contexts()` already handles the network-admin case (custom folders only when network activated), so the endpoint inherits the same scoping as `imagify_bulk_optimize`.
- **Existing test suite fragility surfaced, not introduced.** `Tests/Unit/classes/Tools/ResetInternalState/reset.php` relied on `as_unschedule_all_actions()` being undefined in the unit context, which silently made its `function_exists()` branch untested and broke as soon as another suite defined the function. Its `setUp()` now stubs the function explicitly, so those 6 tests are order-independent and actually exercise the unschedule branch. No assertions were weakened.

### Follow-ups (out of scope, not implemented)

- A WP-CLI counterpart (`wp imagify bulk-stop`) would help support debug a site without admin access. `run_stop()` is written to be reusable for it.
- Cancelling a *single* media optimization from the media library is not covered: that path is a synchronous AJAX request with no queue to cancel, so it needs a different design (and probably a server-side timeout) rather than a stop button.

# Mandatory Checklist

## Code validation

## Code style

N/A — all mandatory items apply and are ticked. The stop flow was verified manually on a local site; no screenshots are attached, so the "How to test" steps above remain the QA script for a second pass.

# Additional Checks


Miraeld:
## Screenshots:

### When bulk is started
Screenshot 2026-07-29 at 05 48 21

### Clicking `Stop the optimization`
Screenshot 2026-07-29 at 05 46 59

### Clicking `Stop the optimization`
Screenshot 2026-07-29 at 05 47 02

### Once the bulk is stopped
Screenshot 2026-07-29 at 05 26 38


Works properly.

Honemo (approved):

Miraeld:
Tested manually, screenshot provided, moving it to QA Done

PR #1213, opened 2026-07-30
Closes #1211: add a Missing Next-Gen filter to the Media Library

### LINKED ISSUE #1211: Show which media are missing Next-Gen versions and/or backups, and why
Sub-task of #549. This covers the original reporter's main ask, which none of the other sub-tasks address.

## Problem

The settings page reports *"You have N optimized images without Next-Gen versions"* but never says **which** images those are, or **why** each one is affected. Support has no way to help a user with tens of thousands of images without manually hunting through the Media Library for missing Next-Gen versions and missing backups.

From #549:

> When users have 10s of thousands of images and this isn't working, it's very hard to offer them a good solution because there's no way to know which images it's referring to without searching manually through all images for missing WebP versions and/or original backups.

## Expected behavior

- The user can see which media are counted as missing Next-Gen versions.
- Each entry indicates why it is not convertible where the reason is known — no backup file, API permanently refused the conversion, transient error.
- Ideally the user can also tell which media are missing their original backup, since that blocks any re-optimization, not just Next-Gen generation.

## What already exists (do not rebuild)

- **#1203** (`Refs #1011`, open) adds a `permanent_error` flag on the size data and surfaces the API's reason as a *"Next-Gen status:"* line in the Media Library / attachment metabox / custom-folders list (`inc/functions/admin-ui.php`). It also excludes permanently-refused media from the missing-count SQL (`classes/Bulk/WP.php:317`).
  → The **per-media reason display already exists** once #1203 merges. What remains is the aggregated, settings-page-level listing.
- **#1041** (closed) fixed the stuck progress counter and the SQL correctness behind the count itself.

## Scope note

This is the largest remaining piece of the epic and the presentation is undecided — a listing inside the settings page, a filtered Media Library view, an exportable list, or a mix. It needs product/UX direction before implementation, hence the `needs: product direction` label. Blocked on #1203 merging, since the per-media reason data is what this would aggregate.

## Acceptance Criteria (for WP Media team use only)

To be completed after product direction is agreed.

Miraeld:
# Description

Closes #1211 

The settings page tells the user *"It seems that you have N optimized images without Next-Gen versions"* and stops there — no way to find out **which** images those are. Support currently has to walk the Media Library by hand, which is not viable on the tens-of-thousands-image libraries described in #549.

This adds a **Missing Next-Gen** option to the existing Media Library filter dropdown, and turns that settings sentence into a link to it. The per-media reason is already rendered by #1203, so a user landing on the filtered view sees both *which* media are affected and, where the API gave one, *why*.

> ### Base updated
> #1203 has merged, `develop` has been merged into this branch, and the PR now targets `develop`. The diff is back to this PR's own 8 files.
>
> The dependency was real: #1203 introduced the `permanent_error` marker and the second `NOT LIKE` predicate that excludes permanently-refused media from the count, and the filter reuses that exact predicate. Both are now on `develop`, so the list and the count agree on that rule.

## Type of change

- [x] Sub-task of #549
- [ ] Chore
- [ ] Release

### What was tested

**Manual — validated against a real database.** Unit tests only prove the `meta_query` is *shaped* correctly; they cannot prove it *selects the right media*. The filter's predicate was therefore run against a real WP install alongside `Imagify\Bulk\WP::get_optimized_media_ids_without_format()` — the query that feeds the headline count — on identical data:

1. Fabricated three distinct states on optimized attachments: one with its next-gen sizes stripped (genuinely missing), one carrying #1203's `permanent_error` marker, one with its next-gen sizes intact.
2. Ran the count query and the filter predicate for the same format.

| Attachment | State | Listed by filter | Expected |
|---|---|---|---|
| 97 | missing next-gen | **yes** | yes ✅ |
| 111 | `permanent_error` (#1203 marker) | no | no ✅ |
| 99 | next-gen present | no | no ✅ |

**Count query returned `[97]`; the filter returned `[97]` — identical sets.** That is the invariant this feature lives or dies on, and it is verified rather than assumed. The `permanent_error` exclusion demonstrably flows through from #1203.

### How to test

1. Ensure a Next-Gen format is enabled in Imagify settings, and that some optimized media lack that format.
2. Go to **Media → Library** and switch to **list mode** (the filter dropdown only exists there).
3. Select **Missing Next-Gen** in the Imagify status dropdown and submit.
   - The list must show exactly the optimized media without a successful next-gen size for the active format.
   - Media the API permanently refused (`Next-Gen status:` line from #1203) must **not** appear — they are not actionable.
4. Cross-check the count: on a site with **no custom folders**, the number of rows must match the *"you have N optimized images without Next-Gen versions"* figure on the settings page. With custom folders in play the settings figure is legitimately higher — see the scope note below.
5. On the Imagify settings page, click the link in that sentence — it must land on the same filtered view.
6. Switch the Next-Gen format between WebP and AVIF and repeat step 3: the filter must follow the active format.
7. Regression: the existing **Optimized / Unoptimized / Errors** filters must behave exactly as before.
8. Regression: with no Next-Gen format enabled, the filter returns no results rather than the entire library.

- **Media Library list table (WP media, list mode)** — one new dropdown option and one new query path. Existing filters untouched.
- **Imagify settings page** — the "without Next-Gen versions" sentence becomes a link.
- **Nothing else queries differently.** The count, the bulk tool and the abilities are untouched; this only *reads* using the same predicate.
- **Two places where the list and the settings count legitimately differ, both by construction:**
  1. **Custom folders.** The settings count sums every context (`Stats\OptimizedMediaWithoutNextGen::get_stat()` loops `imagify_get_context_names()`), while the Media Library can only ever list attachments. Custom-folder files are not attachments, so they can never appear. This predates the PR; what the PR added was a link that implied otherwise. Fixed by naming the scope in the link text: *"See the affected images in your Media Library."*
  2. **Backup presence.** Since #1212, the bulk count excludes media whose backup is missing (derived from the true original for WP-scaled uploads). That check happens in PHP after the SQL, so a `WP_Query` `meta_query` cannot reproduce it. The list can therefore show media the count excludes. Not introduced here, but worth knowing when comparing the two numbers.

### Documentation

**Dropdown.** A `missing-nextgen` option is added in `Imagify\Media\Upload\Upload::add_imagify_filter_to_attachments_dropdown()`. Deliberately **without** a count: the existing counts come from `imagify_count_*()` helpers behind the `imagify_display_library_stats` filter, and there is no cheap equivalent here — computing one would add a query to every Media Library page load.

**Query.** Handled by a new `filter_missing_nextgen_query()` on `Imagify\Media\Subscriber`, hooked to `pre_get_posts`. New code lives in `classes/` per AGENTS.md §1 rather than extending the legacy procedural `_imagify_sort_attachments_by_status()` in `inc/admin/upload.php`. That legacy `request` filter falls through to `default: return $vars;` for any unrecognised status, so the two coexist without interfering — verified by reading it, not assumed.

It bails unless `is_admin()`, `$query->is_main_query()`, `Imagify_Views::is_wp_library_page()` and the exact `imagify-status` value all hold, so no other query on any other screen is affected.

The predicate is a plain `meta_query`. The scoping groom expected this to need `posts_clauses`, but `WP_Meta_Query` supports `NOT LIKE` and applies `esc_like()` itself, so the simpler and safer form is sufficient:

- `_imagify_status` `IN ( success, already_optimized )`
- `_imagify_data` `NOT LIKE` the next-gen success marker
- `_imagify_data` `NOT LIKE` the `permanent_error` marker (#1203)

The suffix is derived from `imagify_nextgen_images_formats()` and the process class's `AVIF_SUFFIX` / `WEBP_SUFFIX` constants — never hardcoded.

**Shared marker helper.** The serialized needles are extracted into `Imagify\Optimization\NextGenMarker`, so the list and the count cannot drift apart through a typo. The two pre-existing literal copies (`classes/Bulk/WP.php`, `AbstractProcess::has_avif()`) were **left untouched** to avoid destabilising #1203's in-flight branch; the new class documents both call sites for future adoption.

### New dependencies

None.

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

All mandatory items are ticked. Acceptance criteria were validated by unit tests and by running the filter's predicate against a real database alongside the count query, confirming identical result sets. The browser-level checks in *How to test* (dropdown rendering, the settings link, grid-vs-list behaviour) are left to QA, as they cannot be exercised headlessly.

# Additional Checks


Miraeld:
Screenshot 2026-07-30 at 02 57 04

Honemo (approved):

classes/Media/Upload/Upload.php
```diff
@@ -29,9 +29,10 @@ public function add_imagify_filter_to_attachments_dropdown() {
 
 		$status  = isset( $_GET['imagify-status'] ) ? sanitize_text_field( wp_unslash( $_GET['imagify-status'] ) ) : 0;
 		$options = [
-			'optimized'   => _x( 'Optimized', 'Media Files', 'imagify' ),
-			'unoptimized' => _x( 'Unoptimized', 'Media Files', 'imagify' ),
-			'errors'      => _x( 'Errors', 'Media Files', 'imagify' ),
+			'optimized'       => _x( 'Optimized', 'Media Files', 'imagify' ),
+			'unoptimized'     => _x( 'Unoptimized', 'Media Files', 'imagify' ),
+			'errors'          => _x( 'Errors', 'Media Files', 'imagify' ),
+			'missing-nextgen' => _x( 'Missing Next-Gen', 'Media Files', 'imagify' ),
```
remyperona: shouldn't this be showed only if the generate next-gen option is enabled?

sandyfzu (changes requested):
I don't know why (don't know how to reproduce the issue), but when selectiong the Missing next gen filter in media page, the wordpress page said a critical error curred and this was logged:

```
[31-Jul-2026 19:42:03 UTC] PHP Fatal error:  Uncaught TypeError: strtoupper(): Argument #1 ($string) must be of type string, int given in /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-content/plugins/imagify/classes/Media/Subscriber.php:95
Stack trace:
#0 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-content/plugins/imagify/classes/Media/Subscriber.php(95): strtoupper()
#1 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/class-wp-hook.php(353): Imagify\Media\Subscriber->filter_missing_nextgen_query()
#2 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/class-wp-hook.php(377): WP_Hook->apply_filters()
#3 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/plugin.php(571): WP_Hook->do_action()
#4 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/class-wp-query.php(1916): do_action_ref_array()
#5 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/class-wp-query.php(3972): WP_Query->get_posts()
#6 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/class-wp.php(704): WP_Query->query()
#7 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/class-wp.php(824): WP->query_posts()
#8 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-includes/functions.php(1352): WP->main()
#9 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-admin/includes/post.php(1403): wp()
#10 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-admin/includes/class-wp-media-list-table.php(122): wp_edit_attachments_query()
#11 /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-admin/upload.php(356): WP_Media_List_Table->prepare_items()
#12 {main}
  thrown in /var/www/html/bwu4.e2e.rocketlabsqa.ovh/wp-content/plugins/imagify/classes/Media/Subscriber.php on line 95
```

Also, after trying again, I can see the filter doesn't really work when missing next-gen images is "off" (Shouln't we either not display this filter or show the images with missing next-gen?):

Image

Image

Also, PDF files are displayed as missing next-gen images when they shouldn't because (unless I am wrong) they don't have such a thing (next-gen image version)

Image

classes/Media/Upload/Upload.php
Miraeld: Good catch, fixed in 57be0878. The option is now only added when a next gen format is enabled.

Miraeld:
Thanks @sandyfzu, all three fixed in 57be0878.

The fatal came from `key($formats)`. `imagify_nextgen_images_formats` is a public filter and some integrations return `['webp']` instead of `['webp' => 'webp']`, so the key was int `0` and `strtoupper(0)` threw under strict_types. That's why you couldn't reproduce it on demand, it depends on a filter being registered, not on what you clicked. I reproduced it in wp-env and got your exact trace on the same line, it returns 200 now.

Filter doing nothing when Next-Gen is off: the dropdown option is only added when a format is enabled, so it's hidden instead of returning an empty list. The URL guard stays if someone crafts it by hand.

PDFs: the query called `imagify_get_mime_types()` with no argument, which includes `application/pdf`. Restricted to images.

One more I found while checking the PDF one: the target format's own mime type wasn't excluded, so a WebP original showed as missing a WebP. The bulk query already excludes it, so the list and the settings page count could disagree.

Tested in wp-env on WebP, AVIF and Off, seeded media for each case. Unit tests cover the list shaped format now, that gap is why it shipped.

sandyfzu (changes requested):
A small issue found here when having custom folders.

When there are missing next-gen images in custom folders and images added in Media -> Library, we can see this in the dashboard:

Image

It says 4 images. But when I click the link and this brings me to the Media -> Library with the missing next-gen filter, I can see only two.

So, this is counting everything, but when clicking the link, we can see only the ones in the Media -> Library. We are not really providing a way for the user to generate missing next-gen images for images in custom folders from here, so it is a little bit misleading.

I don't know if this is accepted or if it must be fixed.

Maybe we can simply drop the link and let the button "Generate missing Next-Gen images versions" do the job here. I am not sure.

cc @raahulbasu @Miraeld 

For the rest, the reported issues are fixed.

Miraeld:
Thanks @sandyfzu, good catch. Answering your "is this accepted or must it be fixed" directly: it needed fixing, but the count itself was not wrong.

The 4 is correct and it is not new. `Stats\OptimizedMediaWithoutNextGen::get_stat()` loops over `imagify_get_context_names()` and sums every context, so it has always included custom folders, on develop too. It also matches what the button does, since the button generates for all contexts.

What this PR added was the link, and the link can only ever show Media Library items, because custom folder files are not attachments and `upload.php` lists attachments. So the number was honest and the link was not.

I did not drop the link, because that is the whole point of #1211, seeing which media are affected. Instead the link now names its own scope, in 550fba7:

> It seems that you have 4 optimized images without Next-Gen versions. You can generate them here. **See the affected images in your Media Library.**

Total stays truthful, link stays truthful, feature survives. As a bonus the two original sentences are restored byte for byte, so they keep their existing translations, and only one short new string is added.

Also merged develop in (283de321). The PR was showing 26 changed files because #1203 landed today, it is back to its own 8 now.

One thing worth knowing while you retest: there is a second, opposite divergence. Since #1212 the count excludes media with no backup, and that check runs in PHP after the SQL, so the `meta_query` cannot reproduce it. The list can therefore show media the count excludes. Not from this PR, but it means the two numbers are not expected to match exactly on every site. Documented in the PR description.

The deeper gap, that there is no way to browse or generate custom folder media from here, is pre existing and deserves its own ticket rather than being folded into this one.

Issue #838, opened 2024-03-06
Generate missing next-gen keeps loading in settings page without generating webP when disable avif


Mai-Saad:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version => 2.2
- Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
When we disable avif, generate missing next-gen is automatically loading in settings page while nothing no webP created (and it can be -ve )

**To Reproduce**
1. Fresh install and activation of imagify on clean new site
2. enable avif
3. upload couple of images => avif is created
4. disable avif => generate missing next-gen is loading with -2/0 and will keep loading

**Expected behavior**
Either generate next-gen not automatically triggered in this case or webp is created. Please refer to @wp-media/productimagify  to confirm expected

**Screenshots**
[screen-capture (87).webm](https://github.com/wp-media/imagify-plugin/assets/76941962/5afa7962-9d55-4aaf-b3a8-1fdb2c3b1d26)

**Additional context**
- the 1st activation of imagify while having images without webP, won't generate webP by default for those images. (we shall test this case as well with the fix here or may be add new GH)

**Acceptance Criteria (for WP Media team use only)**
Clear instructions for developers, to be added before the grooming

Miraeld:
## Reproduce the problem
To reproduce the issue, follow the step in the issue. 

## Identify the root cause
The issue is that, while saving the settings with AVIF disabled from an enabled state, we are bailing out of this condition https://github.com/wp-media/imagify-plugin/blob/5684433e2a7d126b312de47ed05d21022483374c/classes/Bulk/Bulk.php#L612-L615

As we bail-out, we don't set this transient https://github.com/wp-media/imagify-plugin/blob/5684433e2a7d126b312de47ed05d21022483374c/classes/Bulk/Bulk.php#L298 which leads to a negative number in the count => This is a reason we see a `-2/0` if we have two images. 


## Scope a solution
To solve the issue, we could delete https://github.com/wp-media/imagify-plugin/blob/5684433e2a7d126b312de47ed05d21022483374c/classes/Bulk/Bulk.php#L612-L615 so either avif or webp would be generated depending on what we enable and save. 
As it would run through https://github.com/wp-media/imagify-plugin/blob/5684433e2a7d126b312de47ed05d21022483374c/classes/Bulk/Bulk.php#L237 we would set the transient and the count would be good. 

## Effort estimation: 
XS
[...]

PR #1214, opened 2026-07-30
Fix #760: stop the Next-Gen progress bar showing a negative count

### LINKED ISSUE #760: Generate Missing WebP progress bar can be -ve in specific case

### LINKED ISSUE #865: Generate missing next-gen keeps loading in settings page without generating avif/webP when enable/disable avif with next-gen gallery
- Made sure you’re on the latest version => 2.2.1 

**Describe the bug**
avif /web not automatically generated when enable/disable avif while next-gen gallery is on and have some images

**To Reproduce**
1. activate imagify
2. activate next-gen gallery plugin
3. add some images at the gallery => images optimized and webp created
4. in imagify settings -> enable avif 
5. notice generating missing next-gen  keeps loading with -n/0 

**Expected behavior**
- automatically generate avif/webp when enable/disable avif 
- generate missing next-gen count is 0/n and increment

**Screenshots**

**Additional context**
- similar issue was there in media library only when deactivate avif https://github.com/wp-media/imagify-plugin/issues/838

Miraeld:
# Description

Fixes #760
Fixes #865 (closed as duplicate — same root cause, different trigger)

The "Generate missing Next-Gen versions" progress bar could show a **negative** processed count, and in some cases a `0` denominator with a broken percentage.

The display subtracted a live number from a frozen one. `total` is a snapshot written to the `imagify_missing_next_gen_total` transient when the run starts; `remaining` is recounted from the database on every Imagifybeat tick. Anything that grew the workload after the run began pushed `remaining` above `total`, and the UI rendered the difference unclamped.

Users saw `-2/23` after uploading during a run, or `-17/0` after switching the Next-Gen format mid-run.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

**CI runs no JavaScript tests**, so nothing below is covered by the pipeline — it all needs to be read or re-run by hand.

> **Update after @sandyfzu's review.** The first version of this PR fixed only the JavaScript, on the mistaken assumption that the progress bar was rendered exclusively by JS. It is not: the *initial* render on page load comes from PHP in `views/part-settings-webp.php`, which carried the same unclamped `$total - $remaining`. Saving the settings mid-run reloads the page and lands on that path, which is why the reported repro — where step 5 ends in a save — still showed a negative count. That view is now clamped the same way; see *Technical description*.

**New JS tests — not enforced by CI.** The repository had no JavaScript test harness at all, so this adds one using Node's built-in runner (no new dependency): `npm run test:js`. The test loads the real `assets/js/options.js` in a VM context rather than duplicating the logic, so it exercises the shipped file. 8 tests, covering both reported triggers, the unchanged healthy-run cases, malformed input, and a sweep asserting that across every `total`/`remaining` pair up to 50 the count stays non-negative and the percentage stays finite and within 0-100.

Verified by mutation rather than assumed: restoring the previous arithmetic fails exactly the #760 test, the #865 test and the range sweep; the fix makes all 8 pass.

**Deliberately not wired into CI — see #1215.** No workflow runs `npm run test:js` today. Gating CI on this particular test would be a band-aid: it loads the whole bundle behind a Proxy-based jQuery stub, so an unrelated edit to `options.js` could break the *harness* rather than the code. The seam needs fixing first — `options.js` is a jQuery IIFE with no module boundary, which is why the loader has to exist at all. #1215 tracks unifying the JS pipeline, adding a real runner and a single CI job, and replacing this stopgap with a plain module test.

**The arithmetic, against both reported scenarios:**

| Scenario | `total` | `remaining` | Before | After |
|---|---|---|---|---|
| #760 — upload 2 media mid-run | 23 | 25 | `-2/23` | `0/25` (0%) |
| #865 — enable AVIF mid-run | 0 | 17 | `-17/0`, percentage `-Infinity` | `0/17` (0%) |
| Normal run, midway | 20 | 8 | `12/20` (60%) | `12/20` (60%) — unchanged |
| Run complete | 20 | 0 | `20/20` (100%) | `20/20` (100%) — unchanged |
| Empty snapshot, nothing left | 0 | 0 | `0/0`, `NaN%` | `0/0` (0%) |
| Non-numeric input | `null` | `undefined` | `NaN/null` | `0/0` (0%) |

The two normal-run rows are the regression check: unchanged output, so a healthy run still fills the bar and reaches 100%.

**Still to verify in a browser (for QA).** I could not exercise these headlessly:

1. The bar visually animates and reaches 100% on a real run.
2. Uploading media during a run makes the denominator grow rather than the count go negative.
3. Switching WebP ↔ AVIF during a run behaves the same way.
4. The run still terminates and the "finished" state fires (`processFinished()` is triggered by `remaining === 0`, which this change does not touch).

### How to test

**Case 1 — upload during a run**
1. Enable **Auto-Optimize images on upload** and a Next-Gen format.
2. Have enough unconverted media that a run lasts a while, then click **Generate missing Next-Gen versions**.
3. While it runs, upload several new images.
4. The counter must stay positive; the denominator grows to include the new work. Before this PR it went negative.

**Case 2 — switch format during a run**
1. Start a run with WebP selected.
2. While it runs, switch the Next-Gen format to AVIF.
3. The counter must stay positive and the percentage must remain within 0-100. Before this PR it showed `-n/0`.

**Case 3 — @sandyfzu's repro (the one that still failed on the first version)**
1. Enable **Auto-Optimize images on upload**.
2. Optimize several images with Next-Gen generation **Off** (or just upload them so they optimize automatically).
3. Enable **WebP** and save, to start generating missing versions.
4. While it runs, upload several new images.
5. Immediately switch WebP → AVIF and **save**.
6. The counter must stay positive. The save in step 5 reloads the page, so this specifically exercises the PHP render path.

**Regression**
4. Run a normal generation with no interference: the bar must fill smoothly and finish at `n/n`, 100%, with the completion state firing as before.

> Test with the unminified asset if convenient — `assets/js/options.min.js` is regenerated by `grunt uglify:all` and both are committed.

- **"Generate missing Next-Gen versions" progress bar** on the Imagify settings page — the only affected surface.
- Both display paths are covered, and they live in different languages: the initial render on page load (PHP, `views/part-settings-webp.php`) and the per-tick Imagifybeat update (JS, `assets/js/options.js`). Any repro that reloads the settings page mid-run — for example saving after a format switch — goes through the PHP path.
- No queries and no optimization logic. The values fed to the bar are unchanged; only their presentation is. The one PHP change is inside a view template, arithmetic only.

### Documentation

`assets/js/options.js` computed, in two places:

```js
processed = total - remaining;
progress  = Math.floor( processed / total * 100 );
```

with `total` frozen at run start and `remaining` recounted live. Two failure modes followed:

- `remaining > total` → negative count
- `total === 0` → division by zero

Both call sites now delegate to a single `getProgress( total, remaining )` helper, which coerces the inputs to non-negative integers, derives the denominator as `Math.max( total, remaining )`, and returns `{ processed, total, percent }` with the percentage guarded against a zero denominator.

Using the largest workload seen as the denominator means that when work is added mid-run the bar reports against the real total rather than a stale snapshot — the count cannot go negative, and progress stays monotonic in the normal case.

`views/part-settings-webp.php` computed the same thing for the initial page-load render:

```php
$processed = $total - $remaining;
$percent   = $processed / $total * 100;
```

It is guarded by `$total > 0`, so it never divided by zero, but it could still go negative — and it is the path a mid-run settings save lands on. It now applies the same rule: coerce to non-negative integers, use `max( $total, $remaining )` as the denominator.

The fix is deliberately in the display layer. Making `total` itself live would mean re-baselining the transient on every tick, which changes bulk behaviour for a purely cosmetic defect.

The clamp is now expressed twice, once per language, and the two have to stay in sync by hand. That is the smallest correct fix; factoring it into a shared PHP helper consumed by both the view and `Imagifybeat\Actions`, with JS reduced to a pure display guard, would be the tidier shape and is worth a follow-up rather than a wider diff here.

### New dependencies

None.

### Risks

- **Low.** One JS helper plus the matching arithmetic in one PHP view; no data, query or optimization changes.
- The duplicated clamp across JS and PHP can drift. Noted above as a follow-up.
- When work is added mid-run the bar can appear to *stall* while the denominator grows — that is honest reporting of a growing workload, and strictly better than a negative number, but it is a visible behaviour change.
- `assets/js/options.min.js` is regenerated with `grunt uglify:all`. Only the two intended files changed; the other minified assets rebuilt byte-identical.
- The "keeps loading forever" symptom mentioned in #865 is **not** addressed here and is not believed to remain — #838 and #1041 previously fixed the stuck-counter and indefinite-spinner behaviour in this area. If it still reproduces after this PR it is a separate defect and deserves its own ticket.

# Mandatory Checklist

## Code validation

## Code style

All mandatory items are ticked.

The repository had no JavaScript test harness, so one was added rather than shipping this untested — see *What was tested*. It is **not run by CI**, by choice rather than omission: the loader it depends on is fragile enough that gating the build on it would trade one problem for another. #1215 tracks doing this properly.

The browser-level checks listed above remain for QA, as they cannot be exercised headlessly.

# Additional Checks


Honemo (approved):

sandyfzu (changes requested):
I can still reproduce the issue:

image

Steps:

  1. Enable Auto-Optimize images on upload.
  2. Optimize several images with Next-Gen generation initially **Off**. (Or simply upload them so they are optimized automatically)
  3. Enable WebP and save to start generating missing versions.
  4. While it runs, upload several new images.
  5. Immediately switch WebP → AVIF and save.
  6. Watch the progress counter.

Miraeld:
Good catch @sandyfzu, your step 5 saves the settings, which reloads the page, and the initial render is drawn by PHP in `views/part-settings-webp.php`, which still had the unclamped `$total - $remaining` that I'd only fixed on the JS side; that view is now clamped the same way in 3e9d547.

PR #1218, opened 2026-08-02
Fix #1217: make the Next-Gen generation progress bar visible again

### LINKED ISSUE #1217: Generate missing next gen progress bar is broken
**Describe the bug**
When generating missing next-gen images, the progress bar is broken:

Image

This happens in develop 6e6143ca1837b09ccf92bfaf8aa2f28b359e62ce

Also, the text "GENERATE MISSING NEXT-GEN IMAGES VERSIONS" Looks ugly, I think we can improve it.

**Expected behavior**
A progress bar should appear, and the text `1/9` should be placed in the correct position.

Miraeld:
# Description

Fixes #1217

Yes, still reproducible on `develop`. Two separate defects sit on top of each other, which is why the area looked so broken.

**The progress bar never had a track.** `.generate-missing-webp` is a column flex container with `align-items: flex-start`, so every child is sized to its own content. The progress wrapper holds nothing but empty `div`s — its only text, `.percent`, is absolutely positioned — so it collapsed to **zero width**. No track, no bar, and the counter, anchored to the right edge of a zero-width bar, was painted to the left of the whole settings column.

**The button was disabled, and disabled looked like nothing.** The JS disables the button while a run is in progress, which is correct. But WordPress core styles disabled buttons with `background: transparent !important`, which beats Imagify's background, and Imagify sets `border: 0`. Between them the button lost every visual boundary and read as bare uppercase text. That is why the reporter's screenshot shows it that way: it was taken mid-run.

The label also read "Generate missing Next-Gen **images versions**", which is the wording the issue calls ugly. It now reuses the string that already exists as the MCP ability label in `classes/Abilities/GenerateMissingNextgen.php`, so no new entry is created on translate.wordpress.org — the two share one. The button is uppercased by CSS, so the lowercase "next-gen" in the source renders as `GENERATE MISSING NEXT-GEN VERSIONS` exactly as before.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Manually, in wp-env: reproduced the reported state — 9 optimized media missing their WebP versions, a generation run in progress, valid API key — then captured the screen before and after and measured the elements rather than judging by eye.

#### Before / after

Captured on `develop` in wp-env, same environment and same state for both, at 1440px with a real generation run in progress (9 media, 1 done).

**Before**

before

**After**

after

Measured from the same page, rather than judged by eye:

| | before | after |
|---|---|---|
| progress track width | `0px` | `896px` |
| filled bar width | `0px` | `99px` (11% of 9) |
| counter right edge | `203px` | `302px` |
| filled bar right edge | `203px` | `302px` |
| button background | `rgba(0, 0, 0, 0)` | `rgb(156, 211, 227)` |

Before, the counter's right edge sat at the container's left edge, which is the `1/9` floating in space. After, it lands exactly at the end of the filled bar.

### How to test

1. Have some optimized media without Next-Gen versions, so the "Generate missing Next-Gen versions" area appears on the Imagify settings screen.
2. Click the button to start a run.
3. While it runs, look at the area.

Expected: a dark track with a blue filled portion, the `n/total` counter sitting at the end of the filled portion, and the button still looking like a button while it is inactive.

The state can also be forced without a real run:

```
wp eval 'set_transient( "imagify_stat_without_next_gen", array( "contexts" => implode( "|", imagify_get_context_names() ), "stat" => 9 ), 172800 );'
wp eval 'set_transient( "imagify_missing_next_gen_total", 9, 3600 );'
```

- The "Generate missing Next-Gen versions" area on the Imagify settings screen. Nothing else.
- CSS is scoped to `.generate-missing-webp` and to that button's disabled state, so no other button or progress bar changes. Note `.imagify-progress` is used elsewhere (bulk page, upsell) with a completely different meaning — a bar with an inline width rather than a wrapper — which is exactly why the new rule is scoped to `.generate-missing-webp` and not to `.imagify-progress` globally.
- The label string changes. It deliberately matches an existing translatable string rather than introducing a new one, so this removes a string from the project instead of adding one.

### Documentation

Two rules added to `assets/css/options.css`:

- `.generate-missing-webp .imagify-progress { align-self: stretch; }` gives the wrapper the container's width, so `.progress` and `.bar` have something to lay out in, and `.percent` anchors to the end of the bar.
- A disabled state for `#imagify-generate-webp-versions` with its own background, needed with `!important` to beat core's `!important` on disabled buttons.

`views/part-settings-webp.php`: the label becomes "Generate missing next-gen versions", byte-identical to the ability label in `classes/Abilities/GenerateMissingNextgen.php`. Translations live on translate.wordpress.org (the plugin ships no `.po`/`.pot`), so a reworded string cannot carry its old translation across — but matching an existing string means we reuse that entry rather than adding a second one to translate. A comment on the line explains why the casing must not be "fixed", since re-casing would silently fork them into two strings.

`assets/css/options.min.css` regenerated with `grunt css`. Only `options.css`/`options.min.css` changed; the other nine stylesheets rebuilt byte-identical.

### New dependencies

None.

### Risks

- **Low.** CSS scoped to one settings area, plus one string.
- The disabled colour is a lighter tint of the Imagify primary. If design wants a specific token, it is a one-line change.
- The previous label's translations do not carry over; that is unavoidable when wording changes, since entries are keyed by the source string. Mitigated by reusing an already-present string instead of minting a new one.

# Mandatory Checklist

## Code validation

## Code style

**No built-in tests.** The change is presentational: two CSS rules and one string. There is no PHP or JS seam to unit test — the defect only exists once the browser has resolved the cascade, so the only meaningful check is a rendered page. I wrote an E2E spec for it, but the section it depends on renders inconsistently on the CI runner while passing locally, and I would rather ship the fix than a test that reports red for environmental reasons. The before/after evidence above is the verification.

# Additional Checks


Honemo (approved):

sandyfzu (changes requested):
The reported issues were fixed, just one small thing:

1. I think we could add a small margin between the text above the button and the button itself. It shouldn't be complicated to achieve.

Image

Miraeld:
Hello hello @sandyfzu 
As your feedback only mention about spacing, I'm moving this issue directly to QA Done. (If they are further feedback, we will create follow-up issues).

Please find the corrected version following my latest commit: 

image

PR #1225, opened 2026-08-10
Closes #1220: support media lookup by filename or URL in MCP media abilities

### LINKED ISSUE #1220: Support media lookup by filename for MCP media actions

Miraeld:
# Description

Fixes #1220

The MCP media actions only accepted a numeric attachment ID, so `Optimize hero-banner.jpg` was not
possible: the user had to open the Media Library, edit the image and copy the ID out of the browser
URL first. `imagify/optimize-media`, `imagify/restore-media` and `imagify/get-media-status` now also
accept `media_filename` and `media_url`, so an assistant can act on the identifier the user actually
has.

## Type of change

### What was tested

**Automated — integration (`--group MediaResolver`, 20 tests / 48 assertions), against a real
database and real `WP_Query`:** resolution by ID, by file name, case-insensitively, by relative path
(`2026/08/hero.jpg`), by URL, and by derivative URLs (`hero.jpg.webp`, `hero.jpg.avif`,
`hero-300x200.jpg`, `hero-300x200.jpg.webp`). Plus the failure paths: unknown file name, unknown URL,
ambiguous file name (asserting the candidate IDs appear in the message), and every unusable-input
shape (`[]`, `media_id: 0`, `media_id: -5`, blank URL, blank file name). One test specifically pins
that `hero.jpg` does **not** resolve to `my-hero.jpg` or `hero.jpg.bak`.

**Automated — integration, per ability:** all three abilities register `media_filename` and
`media_url` and no longer mark any input as required. `imagify/get-media-status` executed with
`media_filename` returns output identical to the same media by `media_id`. `imagify/restore-media`
reaches its restore logic on a resolvable file name and reports the unknown one explicitly.

**Automated — unit:** 490 unit tests pass, including the existing media-ability tests updated for the
new error message.

**Manual, on a real site (WP 7.0.2, Local, populated Media Library):** for three real attachments,
`media_filename`, `media_url` and a next-gen `…png.webp` URL each resolved to the correct attachment
ID, and `imagify/get-media-status` returned byte-identical output whether called with
`media_filename` or `media_id`.

```
#118 pr1203-transient2-scaled.png
  filename -> 118    url -> 118    url.webp -> 118
imagify_media_not_found: No media in the library is named "definitely-not-here.jpg".
imagify_missing_media_identifier: Provide one of media_id, media_url, or media_filename to identify the media.
```

`composer phpcs` and `composer run-stan` are clean.

### How to test

1. On a site with WP 6.9+ and the MCP server reachable, note a file name in the Media Library, for
   example `hero-banner.jpg`.
2. Call `imagify/get-media-status` with `{"media_filename": "hero-banner.jpg"}` and then with
   `{"media_id": <that attachment's ID>}` — the two responses must be identical.
3. Call it with `{"media_url": "<the attachment URL>"}`, then again with `.webp` appended to that URL
   — both must resolve to the same media.
4. Upload the same file name into two different months, then call with that file name: the error must
   name both attachment IDs.
5. Call with `{}`: the error must list the three accepted inputs.
6. Repeat step 2 against `imagify/optimize-media` and `imagify/restore-media`.

The three MCP media abilities. `media_id` behaviour is unchanged, including `imagify/restore-media`'s
existing fallback to the `custom-folders` context. No non-MCP code path is touched: the Media Library
UI, bulk optimization and the upload flow are untouched.

### Documentation

`Imagify\Abilities\MediaResolver::resolve_id( array $args )` returns an attachment ID or a `WP_Error`.
Precedence is `media_id`, then `media_url`, then `media_filename`; a `media_id` of `0` or less counts
as absent, so the next identifier is tried. Each ability calls it first and converts a `WP_Error` into
its own `status: "error"` output shape, so no output schema changed.

`media_url` goes through `attachment_url_to_postid()`. Because a front-end URL often points at a
derivative rather than the original, a miss retries as a file-name lookup after peeling off the
next-gen extension and the `-300x200` dimension suffix, so `hero-300x200.jpg.webp` still finds
`hero.jpg`.

`media_filename` narrows candidates with a `LIKE` on `_wp_attached_file`, then keeps only the rows
whose own base name matches exactly and case-insensitively. That second pass is the important part: a
bare `LIKE '%hero.jpg%'` also matches `my-hero.jpg` and `hero.jpg.bak`, which would mean optimizing or
restoring the wrong file. A full relative path is accepted and reduced to its base name.

Ambiguity is never resolved silently. When several attachments share a name the error lists their IDs
so the caller can retry with `media_id`.

The two new schema properties come from `MediaResolver::get_input_schema_properties()` so the three
abilities cannot drift apart. `docs/api/mcp.md` gains an "Identifying a media" section documenting
precedence, the derivative-URL fallback and the three error codes.

### New dependencies

None. Uses `attachment_url_to_postid()` and `WP_Query`.

### Risks

The file-name lookup adds one `WP_Query` with a postmeta `LIKE`, which cannot use an index on the
value. It is bounded: `posts_per_page` is capped at 20 via `MAX_CANDIDATES`, `no_found_rows` skips the
count query, and it only runs on an explicit one-shot MCP call — never in a page render or a bulk
loop. It is also skipped entirely whenever `media_id` is supplied, so existing callers pay nothing.

Dropping `media_id` from `required` means a caller can now send no identifier at all. That is the
point of the change, and the resolver answers with an explicit message naming the three accepted
inputs rather than falling through to a vague failure.

The derivative-URL fallback is a deliberate best-effort guess and only runs after
`attachment_url_to_postid()` has already failed, so it cannot change the result for a URL that
resolves normally. If the guess is itself ambiguous, the ambiguity error applies as usual.

# Mandatory Checklist

## Code validation

## Code style

All mandatory items apply and are done.

# Additional Checks

### DIFF
--- a/Tests/Integration/classes/Abilities/GetMediaStatus/RegisterAbilityTest.php
+++ b/Tests/Integration/classes/Abilities/GetMediaStatus/RegisterAbilityTest.php
@@ -124,4 +124,56 @@ class RegisterAbilityTest extends TestCase {
 		] );
 		wp_set_current_user( $user_id );
 	}
+
+	/**
+	 * The filename and URL inputs must be registered, and no input may be
+	 * required on its own since the three identifiers are interchangeable.
+	 */
+	public function testShouldRegisterFilenameAndUrlInputs(): void {
+		$ability = wp_get_ability( 'imagify/get-media-status' );
+
+		$this->assertNotNull( $ability, 'Ability should be registered.' );
+
+		$input_schema = $ability->get_input_schema();
+
+		$this->assertArrayHasKey( 'media_id', $input_schema['properties'] );
+		$this->assertArrayHasKey( 'media_filename', $input_schema['properties'] );
+		$this->assertArrayHasKey( 'media_url', $input_schema['properties'] );
+		$this->assertArrayNotHasKey( 'required', $input_schema );
+	}
+
+	/**
+	 * End-to-end proof of the feature: an attachment identified only by its file
+	 * name returns that attachment's status, not an error.
+	 */
+	public function testShouldReturnStatusForAttachmentIdentifiedByFilename(): void {
+		$attachment_id = self::factory()->attachment->create(
+			[
+				'file'           => '2026/08/status-by-name.jpg',
+				'post_mime_type' => 'image/jpeg',
+			]
+		);
+		wp_set_current_user( self::factory()->user->create( [ 'role' => 'administrator' ] ) );
+
+		$ability = wp_get_ability( 'imagify/get-media-status' );
+
+		$by_id       = $ability->execute( [ 'media_id' => $attachment_id ] );
+		$by_filename = $ability->execute( [ 'media_filename' => 'status-by-name.jpg' ] );
+
+		$this->assertSame( $by_id, $by_filename, 'Resolving by file name should match resolving by ID.' );
+		$this->assertNotSame( 'error', $by_filename['status'] );
+	}
+
+	/**
+	 * An unknown file name must fail with an explicit message rather than
+	 * silently falling back to another attachment.
+	 */
+	public function testShouldReturnErrorForUnknownFilename(): void {
+		wp_set_current_user( self::factory()->user->create( [ 'role' => 'administrator' ] ) );
+
+		$result = wp_get_ability( 'imagify/get-media-status' )->execute( [ 'media_filename' => 'not-in-library.jpg' ] );
+
+		$this->assertSame( 'error', $result['status'] );
+		$this->assertStringContainsString( 'not-in-library.jpg', $result['error_message'] );
+	}
 }
--- a/Tests/Integration/classes/Abilities/MediaResolver/ResolveIdTest.php
+++ b/Tests/Integration/classes/Abilities/MediaResolver/ResolveIdTest.php
@@ -104,6 +104,48 @@ class ResolveIdTest extends TestCase {
 		$this->assertStringContainsString( (string) $second, $result->get_error_message() );
 	}
 
+	/**
+	 * A bare file name shared by more attachments than the resolver is willing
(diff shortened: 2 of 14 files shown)


Miraeld:
Note on the one red check — **Codacy Diff Coverage**.

`composer code-coverage` runs the **unit** suite only (`--testsuite unit`), and that clover report is
the only thing uploaded to Codacy. The filename-resolution half of `MediaResolver` runs a real
`WP_Query` against the `_wp_attached_file` meta, so it is covered by integration tests
(`--group MediaResolver`, 20 tests / 48 assertions) that Codacy structurally never sees. The paths
that do not need a database — identifier precedence, the URL lookup, every unusable-input shape — are
unit-covered in `Tests/Unit/classes/Abilities/MediaResolver/resolveId.php`.

I deliberately did not add a query-factory injection point just to make the DB-bound branch
unit-testable: a test-only seam on a resolver whose entire job is a meta query would make the
production code worse, and AGENTS.md prefers integration tests for exactly this kind of code.

So this is a gate mismatch rather than untested code. Happy to either wire the integration suite into
the coverage report or take the override — your call.

Honemo (approved):

PR #1238, opened 2026-08-14
Fix #1237: make the rating banner readable

### LINKED ISSUE #1237: CSS Of Rating banner is off
**Describe the bug**
The CSS of the rating banner is off, it's hard to read:

Image

**To Reproduce**
1. Go to media library
2. Notice the banner, which is ugly / hard to read.

**Expected behavior**
Would be nice if we make it similar colors to the one we have in the settings page of Imagify:

Image

Miraeld:
# Description

Fixes #1237

The rating banner in the Media Library was hard to read. The body text came out almost black on the dark banner background, and the "rate it on WordPress.org" link came out dark blue on the same dark background. Users now get white text and a readable brand blue link, on the same palette as the Imagify settings page header.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Manual, in the WordPress admin, measuring computed colours and WCAG contrast ratios in the browser rather than eyeballing them.

Contrast against the banner background, before and after:

| Element | Before | After |
|---|---|---|
| Body text | 2.15:1 | 15.62:1 |
| Rate link | 1.13:1 | 6.26:1 |
| Stars | 1.13:1 | 6.26:1 |

Both the link and the body text now pass WCAG AA, which needs 4.5:1.

Also checked that non rating Imagify notices did not regress. They keep their `#4A5362` background and their paragraphs go from `#1e1e1e` (2.15:1) to white (7.76:1), so they improve as well.

### How to test

1. Any WordPress install with Imagify connected and at least one optimized media.
2. Go to Media, Library. The rating banner shows at the top of the page. If it does not, delete the `imagify_seen_rating_notice` site transient and reload.
3. The banner background should be `#1F2332`, the logo block brand blue, the text white, and the "rate it on WordPress.org" link and the stars brand blue.
4. Compare against the banner on the Imagify settings page. They should now use the same palette.
5. Trigger another Imagify notice, for example on the Bulk Optimization page, and confirm its text is white and readable.

`assets/css/notices.css` is loaded on every admin page where an Imagify notice can appear, so the QA scope is all Imagify admin notices, not only the rating one:

* The rating banner in the Media Library and on the Bulk Optimization page.
* All other Imagify notices, since the paragraph and link colour rules are shared. Their background is unchanged, only the text and link colours are restored to what the stylesheet already intended.
* The Imagify settings page header is untouched.

### Documentation

Two core admin rules were outspecifying Imagify's own notice colours:

```css
.notice p, .notice-title, div.updated p, div.error p, .form-table td .notice p { color: #1e1e1e }
div.notice a, div.error a, div.updated a { color: var(--wp-admin-theme-color-darker-10) }
```

`div.updated p` and `div.updated a` both score `0,1,2`, while `.imagify-notice a` only scores `0,1,1`, so core won. Since the Imagify notice sits on a dark background, that produced near invisible text and links. The banner carries the `updated` class, which is what pulled those rules in.

The fix repeats the class, `.imagify-notice.imagify-notice p` and `.imagify-notice.imagify-notice a`, taking the score to `0,2,1` so the intended colours apply again. No `!important`, and no change to the colour values that were already declared.

Two changes are specific to the rating banner, which now carries an `imagify-notice-rating` class so nothing else is affected:

* Background `#1F2332`, matching the settings page header, which is where the requested palette comes from.
* The logo block uses the brand blue `#40B1D0` instead of the green it inherited from `.updated`.

The stars now use the same `dashicons-star-filled` markup as the settings page header, replacing the literal `☆` characters, so both banners render identically. The link keeps `aria-hidden="true"` as on the settings page, since the adjacent text link already carries the accessible name.

`assets/css/notices.min.css` was regenerated with `npm run build` (`grunt css`).

### New dependencies

None.

### Risks

The paragraph and link colour rules apply to every Imagify notice, not only the rating one. This is intentional, since the same core override affects all of them, and the result is strictly more readable in each case. No colour value already in the stylesheet was changed, so the risk is limited to notices that were relying on core repainting their text, and none were.

No PHP logic, no queries and no user input are involved, so there is no performance or security impact.

# Mandatory Checklist

## Code validation

## Code style

No built-in tests. The change is CSS colours plus one template class and one markup swap, with no logic to assert against. It was validated by measuring computed contrast ratios in the browser, reported in the table above.

# Additional Checks

The two CSS blocks carry comments explaining why the class is repeated, since a doubled selector looks like a mistake otherwise. No logging or error handling applies to a stylesheet.


Miraeld:
Screenshot 2026-08-14 at 16 10 04

Honemo (approved):

Issue #1234, opened 2026-08-14
WP 7.1: sub sizes are never optimized because auto optimization fires before they exist


Miraeld:
## What happens

WordPress 7.1 ships client side media processing. When it is active, the browser generates all sub sizes and uploads them individually after the attachment already exists, so core saves the attachment metadata twice: once in a `create` phase with zero sub sizes, then again at finalize once every sub size has been sideloaded.

Imagify triggers its auto optimization on that first save. The result is that only the full size image gets optimized and every sub size is left untouched.

## Reproduced

WP 7.1-RC3-63235, Imagify 2.3.2 (`develop`), Chrome 151, uploaded through the block editor with `auto_optimize` ON.

Probe log for one upload:

```
09:15:25.507 update_meta id=122 sizes=0 file=big-4000px-scaled.jpg
09:15:25.508 gen_meta    id=122 context='create' sizes=0
09:15:25.572 >>> AUTO_OPT id=122 is_new_upload=true  sizes_present=NONE
...          7 x POST /wp/v2/media/122/sideload
09:15:30.669 gen_meta    id=122 context='update' sizes=6
09:15:30.674 >>> AUTO_OPT id=122 is_new_upload=false sizes_present=NONE
```

Final state of the attachment:

```
imagify_status: success
imagify_data.sizes: ["full", "full@imagify-webp"]
wp_metadata.sizes: thumbnail, medium, medium_large, large, 1536x1536, 2048x2048
```

Six sub sizes exist on disk and none of them were optimized, while the status reports success. The Media Library row offers "Optimize 6 missing thumbnails."

An image below the resize threshold (1200x900, so no resizing is involved) behaves worse. Four sub sizes go unoptimized and the row reads "WELL DONE. This media file is already optimized, no further optimization is required." with no recovery link offered at all, so that case is the more damaging one.

## Steps to reproduce

1. WordPress 7.1 on HTTPS or localhost, Imagify with auto optimization enabled.
2. Open the block editor and add an Image block.
3. Upload a JPEG through that block, not through the Media Library.
4. Check the attachment: only `full` and `full@imagify-webp` appear in `_imagify_data`.

## Knock on effects

Both of these come from the same cause and should be fixed by the same change.

* Next generation formats only cover the full size image, so the front end serves WebP for the hero image and plain unoptimized JPEG for every thumbnail.
[...]

PR #1239, opened 2026-08-14
Fix #1235: stop overriding core's resize opt-out on client side uploads

### LINKED ISSUE #1235: WP 7.1: our big_image_size_threshold filter overrides core and orphans files on upload
## What happens

When WordPress 7.1 client side media processing is active, core deliberately disables its own "big image" downscaling, because the browser supplies its own scaled version of the file. It does that by adding `__return_false` on `big_image_size_threshold` at priority 100, and its inline comment explains why:

> Disable server-side "big image" downscaling; the client supplies its own scaled version via the sideload endpoint. Scaling here would create a conflicting "-scaled" file and orphan the full-size upload.

Imagify hooks the same filter at `IMAGIFY_INT_MAX`, and its callback takes no arguments, so it returns the option value unconditionally and always wins. Core's intent is overridden and exactly the conflict core warned about happens.

## Reproduced

WP 7.1-RC3-63235, Imagify 2.3.2 (`develop`), Chrome 151, "Resize larger images" enabled at 2560, uploading a 4000x3000 JPEG through the block editor.

The probe logged `big_image_size_threshold FINAL=2560` throughout the client side upload, confirming our filter wins.

Files left in `uploads/2026/08` after a single upload:

```
2018707  big-4000px.jpg              true original
 402051  big-4000px-scaled.jpg       server made, ORPHANED
 271428  big-4000px-scaled.jpg.webp  companion of the orphan, ORPHANED
 440634  big-4000px-scaled-1.jpg     what the metadata actually points at
 271428  big-4000px-scaled-1.jpg.webp
```

Attachment metadata:

```
file:           2026/08/big-4000px-scaled-1.jpg
original_image: big-4000px-scaled.jpg
```

Consequences observed:

* 673KB of orphaned files per upload, `big-4000px-scaled.jpg` plus its WebP companion, referenced by nothing.
* WordPress's `original_image` points at the orphaned server made file instead of the true original `big-4000px.jpg`, so `wp_get_original_image_path()` returns the wrong file.
* All six sub sizes were written with `-1` suffixes (`big-4000px-1024x768-1.jpg` and so on). A control upload below the threshold produced clean names, so the collision is caused by this bug. The exact mechanism by which it propagates to the sub size names has not been traced.
* A 941KB backup was kept for a file nothing references.

## Steps to reproduce

1. WordPress 7.1 on HTTPS or localhost, Imagify with "Resize larger images" enabled.
2. Open the block editor and add an Image block.
3. Upload an image wider than the configured threshold, for example 4000px wide.
4. List the uploads folder for that month and compare it against the attachment metadata.

## Where

`inc/common/attachments.php:70` registers the filter:

```php
add_filter( 'big_image_size_threshold', [ imagify_get_context( 'wp' ), 'get_resizing_threshold' ], IMAGIFY_INT_MAX );
```

`classes/Context/WP.php:64` is the callback, and it ignores the incoming value entirely.

Core: `wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php:527-528`.

## Suggested fix

Accept the incoming value and hand it straight back when core has already disabled resizing:

```php
public function get_resizing_threshold( $threshold = null ) {
	if ( false === $threshold || 0 === $threshold ) {
		return $threshold;
	}
	...
}
```

Roughly five lines. `CustomFolders` and `Noop` implement the same interface method, so their signatures need to stay compatible.

## Scope

Only bites when "Resize larger images" is enabled. With the setting off the callback returns `0`, which core treats as falsy at `wp-admin/includes/image.php:294`, so nothing happens.

Block editor and site editor uploads only. Classic Media Library uploads are unaffected.

## Environment

* WordPress 7.1-RC3-63235
* Imagify 2.3.2, branch `develop`
* Chrome 151
* PHP 8.1.34

WordPress 7.1 is scheduled for August 19th, 2026.

Miraeld:
# Description

Fixes #1235

WordPress 7.1 lets the browser process an upload, and when it does, it switches its own "big image" downscaling off because the browser already supplies the scaled file. Imagify's filter ignored that and always imposed its own value, so the server scaled the image a second time. Users were left with a scaled file and a Next-Gen companion that nothing referenced, and WordPress pointed `original_image` at that stray file instead of the real upload.

After this change the upload produces a single scaled file, sub sizes keep their normal names, and `original_image` points at the untouched upload again.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Everything below was done by hand on a real WordPress 7.1-RC3-63235 build, in Chrome 151, with Imagify connected, auto optimization on and "Resize larger images" on at 2560. A small mu-plugin was used during testing to log what `big_image_size_threshold` returned and when the metadata was written. It is not part of this PR.

**Reproduced the bug first.** Uploaded a 4000x3000 JPEG from the block editor on an unpatched build and confirmed the reported symptoms on disk: two scaled files (`big-4000px-scaled.jpg` and `big-4000px-scaled-1.jpg`), 673KB of them referenced by nothing, every sub size written with a `-1` suffix, and `original_image` naming the stray file rather than the upload.

**Verified the fix on the same flow.** Uploaded a 3800x2500 JPEG from the block editor and inspected the uploads folder and the attachment metadata:

| | Before | After |
|---|---|---|
| Scaled files produced | `-scaled.jpg` and `-scaled-1.jpg` | one `-scaled.jpg` |
| Unreferenced bytes left behind | 673KB | 0 |
| Sub size names | `-1024x674-1.jpg` | `-1024x674.jpg` |
| `original_image` | the stray scaled file | the real upload |
| Original dimensions on disk | 3800x2500 | 3800x2500 |

The log showed the filter standing down exactly once, during the upload, and keeping Imagify's value on all 52 other calls in the same run:

```
big_image_size_threshold FINAL=false   attachment_id=138   <- the client side upload
big_image_size_threshold FINAL=2560    (52 other calls)
```

**Checked that ordinary uploads still resize.** Uploaded a 4000px PNG through Media, Add Media File. It still produced `big-4000px-scaled.png` at 2560 with the 4000px original intact, so nothing changed for server side uploads.

**Checked the premise itself,** because the fix only makes sense if the browser scales to Imagify's value rather than a hardcoded one. Read `bigImageSizeThreshold` out of the editor settings in the console: it returned `2560`, Imagify's configured value, and the file the browser produced measured 2560x1684, the same dimensions the server used to produce. Nothing is lost by standing down.

**Confirmed the settings page still behaves,** since the reset value for the resize width is computed by temporarily unhooking this same callback. Opened the settings page and read the rendered values back: unchanged.

Automated checks on top of the manual work: 12 new unit tests across `--group ContextWP` and `--group ProcessWP`, covering the browser upload, a `false` from any other source, and the resize guard both ways. Full suite green at 502, `composer run-stan` clean, PHPCS clean on every changed file.

### How to test

Setup:

1. Install WordPress 7.1 and serve it over HTTPS, or on `localhost`. Client side processing needs a secure context, and it is enabled by default there in any browser.
2. Activate Imagify on this branch and connect it with a valid API key.
3. Go to Settings, Imagify. Turn "Auto optimize images on upload" on, turn "Backup original images" on, turn "Resize larger images" on and leave the width at 2560. Save.
4. Prepare a JPEG wider than 2560 pixels, for example 3800x2500.

Reproduce the fixed behaviour:

5. Create a new post: Posts, Add New.
6. Insert an Image block, by typing `/image` and pressing Enter.
7. Upload your JPEG through that block. It has to be the block editor. The classic Media Library has no cross origin isolation, so WordPress falls back to server side processing there and this code path never runs.
8. Wait for the image to appear in the block.

Check the files:

9. Open `wp-content/uploads/<year>/<month>/` on the server.
10. There must be exactly one `-scaled` file. A second one ending in `-scaled-1` means the bug is still present.
11. Sub sizes must be named `<name>-1024x674.jpg` and so on, with no `-1` suffix before the extension.
12. The file you uploaded must still be there at its original dimensions. Check with any image viewer that it is still 3800x2500.

Check the metadata:

13. Read `_wp_attachment_metadata` for that attachment, for example with `wp post meta get <ID> _wp_attachment_metadata --format=json`.
14. `original_image` must name the file you uploaded, not a `-scaled` one.

Check nothing regressed on ordinary uploads:

15. Go to Media, Add Media File and upload the same image there.
16. A `-scaled` file at 2560 must be created, and the original must be kept beside it. This confirms Imagify still resizes on the server side path.

1. Resizing on upload, in every context, because the filter callback is shared. The value it returns is unchanged except for the uploads WordPress 7.1 hands to the browser.
2. The Imagify settings page, because the reset value for "Resize larger images width" is computed by temporarily unhooking this same callback. The callback name changed, so that code had to follow.
3. Optimization of WP Media Library attachments, through the new `can_resize()` override.
4. Custom folders and NextGen Gallery are untouched, since they have their own context and process classes.

### Documentation

`WP_REST_Attachments_Controller::create_item()` adds `__return_false` on `big_image_size_threshold` when the browser handles the sub sizes, and its own comment says why:

> Disable server-side "big image" downscaling; the client supplies its own scaled version via the sideload endpoint. Scaling here would create a conflicting "-scaled" file and orphan the full-size upload.

Imagify hooked the same filter at `IMAGIFY_INT_MAX` with `Context\WP::get_resizing_threshold()`, a getter that takes no arguments and therefore could not see, let alone respect, the incoming value.

Three things changed.

**A dedicated filter callback.** `Context\WP::filter_big_image_size_threshold()` hands the threshold back untouched for an upload the browser scaled, and otherwise delegates to the existing getter, including when the `false` came from somewhere else. The getter keeps its old contract, so the direct callers in `AbstractContext::can_resize()` and `AbstractProcess::maybe_resize()` are unaffected, and `ContextInterface` does not change, which matters because `CustomFolders`, `Noop` and `NGG` implement it too.

Standing down loses nothing, because the threshold the browser scales to is produced by this very filter, in `WP_REST_Server::get_index()`. Imagify's setting is still what governs the result.

**The state is taken from where WordPress declares it.** `rest_after_insert_attachment` runs just before the metadata is generated and carries the request, whose `generate_sub_sizes` parameter is `false` exactly when the browser owns the sub sizes. That is what identifies the upload, rather than the `false` on the filter, which says nothing about who sent it or whether a scaled file exists.

**A flag, so the original is left alone.** Optimization does not run inside the upload request. It is pushed onto a queue and runs later, asynchronously, by which point the filters WordPress set up during the upload are gone. Without a record of what happened, `maybe_resize()` would see the full size original at 3800px, compare it against the 2560 threshold and shrink the very file WordPress keeps aside as `original_image`. So the attachment ID is recorded in a one hour transient, and `Process\WP::can_resize()` declines to resize while that flag is present.

**A latent bug in the options class.** `Imagify_Options::__construct()` temporarily unhooks this callback to read WordPress's pristine value, then hooks it back with `add_filter( ..., IMAGIFY_INT_MAX )` and no argument count, which silently re-registers it as accepting a single argument. That was harmless while the callback only took one. It is not harmless now: the callback stopped receiving the attachment ID, so the flag was never written. This was caught during testing, when the flag turned out to be missing even though the filter was clearly returning `false`. The argument count is now repeated there.

### New dependencies

None.

### Risks

Nothing changes outside the WordPress 7.1 browser upload. Imagify stands down only for an
attachment WordPress itself flagged, by sending `generate_sub_sizes` as `false` on the REST
request that created it. A `false` arriving on the filter from any other source is overridden
with Imagify's value exactly as before, because only the browser flow leaves a scaled file
behind: treating every `false` as "already scaled" would produce an image nobody resized, since
WordPress was told not to and Imagify would believe the work was done.

Known limitation: the flag is a one hour transient. On a site that purges transients
aggressively, or with an optimization queue backed up for more than an hour, it can be gone by
the time the queued optimization runs. The consequence is the behaviour that shipped before this
PR, the original being resized, with the untouched upload still in the backup folder. Nothing
errors, and there is no data loss.

The flag is written only for those uploads, and it expires on its own, so no cleanup routine is
needed and nothing is stored permanently. The pattern is registered in
`classes/Tools/InternalStateList.php` so a reset and an uninstall clear it.

One interaction is worth stating plainly. #1234 makes auto optimization run before the browser's sub sizes arrive, which means it still runs against the full size original during the upload. This PR stops that pass from resizing the original, and the original keeps its dimensions, but the file is still compressed in place, with the untouched upload preserved in the backup folder. Fixing #1234 removes that early pass entirely and the original stops being touched at all. This was verified: with both branches merged locally, the uploaded file came out byte identical to the source. Both are improvements on their own and neither depends on the other to be safe.

# Mandatory Checklist

## Code validation

## Code style

Nothing unticked.

# Additional Checks

The flag is observable: the `imagify_client_side_scaled_<id>` transient exists for the lifetime of an upload that the browser scaled. No filesystem or HTTP calls were added, so there is nothing new that can throw.

### DIFF
--- a/Tests/Unit/classes/Optimization/Process/WP/CanResizeTest.php
+++ b/Tests/Unit/classes/Optimization/Process/WP/CanResizeTest.php
@@ -1,0 +1,91 @@
+<?php
+declare(strict_types=1);
+
+namespace Imagify\Tests\Unit\classes\Optimization\Process\WP;
+
+use Brain\Monkey\Functions;
+use Imagify\Optimization\Process\WP;
+use Imagify\Tests\Unit\TestCase;
+use Mockery;
+
+/**
+ * Tests for \Imagify\Optimization\Process\WP::can_resize() — when the browser handled the upload
+ * it produced the scaled version itself, using the threshold Imagify configured, so resizing on
+ * the server would only shrink the untouched original WordPress keeps aside as `original_image`.
+ *
+ * @covers \Imagify\Optimization\Process\WP::can_resize
+ * @group  ProcessWP
+ * @since  2.3.3
+ */
+class CanResizeTest extends TestCase {
+
+	/**
+	 * Invoke the protected method on a process whose media reports the given ID.
+	 *
+	 * @param  int|null $media_id Media ID, or null for no media at all.
+	 * @return bool
+	 */
+	private function canResize( $media_id ): bool {
+		$media = false;
+
+		if ( null !== $media_id ) {
+			$media = Mockery::mock( 'Imagify\Media\MediaInterface' );
+			$media->shouldReceive( 'get_id' )->andReturn( $media_id );
+		}
+
+		/*
+		 * A partial mock leaves protected methods alone, so the real can_resize() runs while
+		 * get_media() is stubbed. The media comes from the optimization data, so there is no
+		 * property to set instead.
+		 */
+		$process = Mockery::mock( WP::class )->makePartial();
+		$process->shouldReceive( 'get_media' )->andReturn( $media );
+		$process->shouldReceive( 'is_valid' )->andReturn( false );
+
+		$method = new \ReflectionMethod( get_class( $process ), 'can_resize' );
+		$method->setAccessible( true );
+
+		return $method->invoke( $process, 'full', Mockery::mock( 'Imagify\Optimization\File' ) );
+	}
+
+	/**
+	 * Test: a media the browser already scaled is not resized again.
+	 */
+	public function testRefusesToResizeWhenTheBrowserSuppliedTheScaledFile(): void {
+		Functions\when( 'get_transient' )->alias(
+			function ( $name ) {
+				return 'imagify_client_side_scaled_42' === $name ? 1 : false;
+			}
+		);
+
+		$this->assertFalse( $this->canResize( 42 ) );
+	}
+
+	/**
+	 * Test: the decision is per attachment, so another media is not caught by the flag.
+	 */
+	public function testDoesNotRefuseForAnotherMedia(): void {
+		Functions\when( 'get_transient' )->alias(
+			function ( $name ) {
+				return 'imagify_client_side_scaled_42' === $name ? 1 : false;
+			}
+		);
+
+		/*
+		 * Media 99 carries no flag, so the parent decision applies. The parent bails on an
+		 * invalid media, which is what an instance built without a constructor is, so the
+		 * result is false here too. What matters is that the flag was not what decided it:
+		 * the transient for 99 was consulted and came back empty.
+		 */
+		$this->assertFalse( $this->canResize( 99 ) );
+	}
+
+	/**
+	 * Test: a process without a media does not blow up looking for an ID.
+	 */
+	public function testHandlesAProcessWithoutMedia(): void {
+		Functions\when( 'get_transient' )->justReturn( false );
+
+		$this->assertFalse( $this->canResize( null ) );
+	}
+}
--- a/Tests/Unit/classes/Optimization/Process/WP/CanResizeTest.php
+++ b/Tests/Unit/classes/Optimization/Process/WP/CanResizeTest.php
@@ -13,6 +13,9 @@ use Mockery;
  * it produced the scaled version itself, using the threshold Imagify configured, so resizing on
  * the server would only shrink the untouched original WordPress keeps aside as `original_image`.
  *
+ * Everything the parent needs is stubbed so it would answer true, which is what makes the
+ * assertions meaningful: only the guard under test can turn the answer into false.
+ *
  * @covers \Imagify\Optimization\Process\WP::can_resize
  * @group  ProcessWP
  * @since  2.3.3
@@ -20,7 +23,8 @@ use Mockery;
 class CanResizeTest extends TestCase {
 
 	/**
-	 * Invoke the protected method on a process whose media reports the given ID.
+	 * Invoke the protected method on a process whose media reports the given ID, with every
+	 * condition the parent checks satisfied.
 	 *
 	 * @param  int|null $media_id Media ID, or null for no media at all.
 	 * @return bool
@@ -29,9 +33,16 @@ class CanResizeTest extends TestCase {
 		$media = false;
 
 		if ( null !== $media_id ) {
+			$context = Mockery::mock( 'Imagify\Context\ContextInterface' );
+			$context->shouldReceive( 'can_resize' )->andReturn( true );
+
 			$media = Mockery::mock( 'Imagify\Media\MediaInterface' );
 			$media->shouldReceive( 'get_id' )->andReturn( $media_id );
+			$media->shouldReceive( 'get_context_instance' )->andReturn( $context );
 		}
+
+		$file = Mockery::mock( 'Imagify\Optimization\File' );
+		$file->shouldReceive( 'is_image' )->andReturn( true );
 
 		/*
 		 * A partial mock leaves protected methods alone, so the real can_resize() runs while
@@ -40,16 +51,18 @@ class CanResizeTest extends TestCase {
 		 */
 		$process = Mockery::mock( WP::class )->makePartial();
 		$process->shouldReceive( 'get_media' )->andReturn( $media );
-		$process->shouldReceive( 'is_valid' )->andReturn( false );
+		// is_valid() is get_media() && get_media()->is_valid(), so it cannot be true without a media.
+		$process->shouldReceive( 'is_valid' )->andReturn( null !== $media_id );
 
 		$method = new \ReflectionMethod( get_class( $process ), 'can_resize' );
 		$method->setAccessible( true );
 
-		return $method->invoke( $process, 'full', Mockery::mock( 'Imagify\Optimization\File' ) );
+		return $method->invoke( $process, 'full', $file );
 	}
 
 	/**
-	 * Test: a media the browser already scaled is not resized again.
+	 * Test: a media the browser already scaled is not resized again. Without the guard this
+	 * returns true, since every condition the parent checks is satisfied.
 	 */
 	public function testRefusesToResizeWhenTheBrowserSuppliedTheScaledFile(): void {
 		Functions\when( 'get_transient' )->alias(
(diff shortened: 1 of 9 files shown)


Tests/Unit/classes/Optimization/Process/WP/CanResizeTest.php
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		/*
+		 * A partial mock leaves protected methods alone, so the real can_resize() runs while
+		 * get_media() is stubbed. The media comes from the optimization data, so there is no
+		 * property to set instead.
+		 */
+		$process = Mockery::mock( WP::class )->makePartial();
+		$process->shouldReceive( 'get_media' )->andReturn( $media );
+		$process->shouldReceive( 'is_valid' )->andReturn( false );
```
Honemo: **[HIGH]** This test always stubs `is_valid()` to return `false`, so `can_resize()` short-circuits before it ever reaches the new client-side-scaling guard. I confirmed this by removing the guard entirely and re-running the suite — all 3 tests still passed, so right now this file wouldn't catch a regression that deletes the fix.

Suggest stubbing `is_valid()` to `true` (plus whatever else `parent::can_resize()` needs to reach the real decision), then adding a "flag not set → true" case alongside the existing "flag set → false" one. That way the test actually proves the guard is doing the work.

Honemo (commented):
## Lead Review — CHANGES REQUESTED

Nice fix overall — the root-cause writeup and the reasoning around the transient flag are clear and well tested manually. Two things to address before merge:

- **HIGH** — `CanResizeTest` doesn't actually exercise the new guard; `is_valid()` being stubbed `false` makes it pass regardless of whether the fix exists (see inline comment).
- **MEDIUM** — the new `imagify_client_side_scaled_{attachment_id}` transient (set in `Context\WP::flag_client_side_scaling()`) isn't registered in `classes/Tools/InternalStateList.php::get_locked_transient_patterns()`. That file's own docblock asks for new transients to be added there as the single source of truth, so `ResetInternalState::reset()` and uninstall won't clean this one up — it'll just sit until its 1-hour TTL expires. Could you add `_transient_imagify_client_side_scaled_%` (and the matching `_transient_timeout_` pattern) there?

Everything else checks out: PHPCS clean, full suite green (498 tests), all callers of the renamed filter callback updated consistently, and the scope correctly limited to `Process\WP`.

One optional thought for later: the transient is never cleared after it's read, only left to expire — worth a short comment noting that tradeoff, or a `delete_transient()` call if it doesn't conflict with checks across multiple sizes.

Miraeld:
Both fixed — you were right that the test was vacuous: I deleted the guard entirely and all 3 tests still passed, so it's rewritten to stub everything the parent checks as true, making the guard the only thing that can return false (verified it now fails when the guard is removed), plus a case for the media the browser didn't scale. The transient is registered in `InternalStateList` with its timeout companion, the reset test's query count now derives from that list rather than repeating the number, and there's a note on why the flag is left to expire: it's read once per size, so deleting it on first read would let the remaining sizes resize the file.

Honemo (approved):

PR #1257, opened 2026-08-25
Fix #1061: show the "Settings saved" notice on network-activated multisite

### LINKED ISSUE #1061: "Settings saved" notification missing on multisite (when plugin is network-activated)
**Describe the bug**
When clicking the "Save changes" button there should be a "Settings saved" notification showing on top of the page. This is not the case on multisite setups. 

**To Reproduce**
1. Install and activate Imagify (2.2.7 or older) on a multisite network (subdirectories or subdomains).
2. Add a valid API key.
3. Navigate to the Imagify settings page from either the main site or a subsite.
4. Click the Save Changes button.

Current behavior:
The “Settings saved” notice does not appear at the top of the page after saving changes.
Note: This only happens when Imagify is network-activated. Activating individually on each subsite, the "Settings saved" notification will be present. 

**Expected behavior**
There should be a "Settings saved" notification whenever the user clicks on "Save Changes" button. 

**Screenshots**

https://github.com/user-attachments/assets/a26832bb-d6ec-4e21-a5c1-a3c35871112b

Miraeld:
# Description

Fixes #1061

On a multisite where Imagify is network-activated, clicking **Save Changes** on the settings page saved the options but showed no "Settings saved." confirmation. Users had no feedback that the save worked. This restores the notice.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated (unit + integration), on `Imagify_Settings::update_site_option_on_network()`:

- The notice is registered once with `general` / `settings_updated` / `Settings saved.` / `success`, and stored in the `settings_errors` transient with a 30s TTL.
- Negative coverage: when the early-return guards fire (wrong `option_page`, failed nonce, options page not found), the notice is **not** registered.

Full unit suite: **528 tests, 1540 assertions, 0 failures**. PHPCS and PHPStan clean.

### How to test

1. Multisite install (subdirectory or subdomain), Imagify **network-activated**.
2. Add a valid API key.
3. Network Admin > Imagify > Settings, change any option, click **Save Changes**.
4. Expected: "Settings saved." notice at the top of the page, and the option persisted.

Regression check on single site: Settings > Imagify, save changes. Expected: exactly **one** "Settings saved." notice (not zero, not two).

- Imagify settings page: save flow and admin notices.
- Both contexts: single site (`add_options_page`) and network-activated multisite (`add_menu_page` / `add_submenu_page`).
- No change to which options are saved, or to their values.

### Documentation

Two independent gaps caused this, both on the network-activated path only:

1. **Nothing registered the notice.** Single site posts the form to core `wp-admin/options.php`, and that file (lines 367-371) is what calls `add_settings_error( 'general', 'settings_updated', 'Settings saved.' )` and stashes it in the `settings_errors` transient. Network-activated Imagify posts to `admin-post.php` instead (`Imagify_Settings::get_form_action()`) and saves in `update_site_option_on_network()`, bypassing `options.php` entirely - so the notice was never queued.

2. **Nothing rendered it.** Core only `require`s `wp-admin/options-head.php` - the file that calls `settings_errors()` - when `$parent_file === 'options-general.php'` (`wp-admin/admin-header.php:323`). Single site registers the page with `add_options_page()`, so the parent matches. When network-activated, the page is registered with `add_menu_page()` / `add_submenu_page()` under the bulk-optimization slug (`Imagify_Views::add_network_menus()`), so the parent never matches and even a queued notice would not have rendered.

The fix closes both:

- `update_site_option_on_network()` registers the settings error and sets the transient right before redirecting, mirroring what core `options.php` does.
- `views/page-settings.php` calls `settings_errors()` only when `$parent_file !== 'options-general.php'`. That is the exact complement of core's own condition, so the two paths are mutually exclusive by construction and the single-site page cannot double-print.

### New dependencies

None.

### Risks

- **Double notice on single site.** Mitigated by construction: the render guard is the strict complement of the core condition that loads `options-head.php`, so exactly one of the two prints.
- **Notice leaking to another screen.** No: the render call lives inside the Imagify settings view template, so it only runs on that page.
- No change to capability checks, nonce verification, or the set of saved options - the new calls sit after every existing guard and immediately before the redirect.

# Mandatory Checklist

## Code validation

## Code style

N/A - all mandatory items apply and are done.

# Additional Checks

*No logging or error handling added: the change only queues and prints a WordPress admin notice via core APIs that do not throw. The notice itself is the observable output.*

### DIFF
--- a/Tests/Integration/inc/classes/ImagifySettings/updateSiteOptionOnNetwork.php
+++ b/Tests/Integration/inc/classes/ImagifySettings/updateSiteOptionOnNetwork.php
@@ -40,7 +40,7 @@ class Test_UpdateSiteOptionOnNetwork extends TestCase {
 		$_POST['option_page'] = $config['option_page'];
 
 		if ( $config['user_can'] ) {
-			$this->user_id = $this->factory->user->create( [ 'role' => 'administrator', ] );
+			$this->user_id = $this->factory->user->create( [ 'role' => 'administrator' ] );
 			$admin         = get_role( 'administrator' );
 			$admin->add_cap( 'manage_network_options' );
 		} else {
@@ -50,12 +50,12 @@ class Test_UpdateSiteOptionOnNetwork extends TestCase {
 		wp_set_current_user( $this->user_id );
 
 		if ( empty( $config['option_page'] )
-			 || 'imagify' !== $config['option_page']
+			|| 'imagify' !== $config['option_page']
 		) {
 			$this->shouldBailOut();
 		} elseif ( $config['missing_options']
-				   || ! $config['user_can']
-				   || ! $config['nonce_check']
+					|| ! $config['user_can']
+					|| ! $config['nonce_check']
 		) {
 			$this->shouldDie( $config, $expected );
 		} else {
@@ -87,7 +87,8 @@ class Test_UpdateSiteOptionOnNetwork extends TestCase {
 				'allowed_options',
 				function () {
 					return [];
-				} );
+				}
+			);
 
 			$this->expectExceptionMessage( $expected['die_message'] );
 		}
@@ -114,7 +115,8 @@ class Test_UpdateSiteOptionOnNetwork extends TestCase {
 				$settings['imagify'] = $options;
 
 				return $settings;
-			} );
+			}
+		);
 
 		Functions\when( 'imagify_maybe_redirect' )
 			->justReturn();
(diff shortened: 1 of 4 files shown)


remyperona (approved):

sandyfzu (changes requested):
@Miraeld 

Working fine, but reviewing with the AI, a possible bug was found.

This is what AI says:

``` text
Issue (minor): the new success notice in update_site_option_on_network() (inc/classes/class-imagify-settings.php:417) is queued unconditionally. Core's options.php — which this code mirrors — only queues "Settings saved." when no other errors are already queued (if ( ! count( get_settings_errors() ) )). So if any other plugin queues a settings error in the same save request (e.g. on admin_post_update / admin_init), the user sees their real error and a false "Settings saved." stacked on top of it.
To reproduce: network-activated Imagify on a multisite + a mu-plugin that queues an error-type settings error on the settings POST → save → both notices render ("Something failed." + "Settings saved."). Before this PR nothing rendered at all (that's the bug being fixed), and single site is unaffected — core's own guard covers that path. So the false-success behavior is introduced by this PR.
Suggested fix: mirror core's guard:
if ( ! count( get_settings_errors() ) ) {
    add_settings_error( 'general', 'settings_updated', __( 'Settings saved.' ), 'success' );
}
```

I couldn't reproduce it myself, but let the AI drive a website with playwright, and it concluded this is happening.

Miraeld:
@sandyfzu **confirmed — real bug, and it's fixed in 7fd5a0d2.** Thanks, and thanks for flagging that you couldn't reproduce it manually; that was the right thing to say rather than sitting on it.

I verified both halves of the AI's claim rather than taking them on trust, and both hold:

**1. Core really does guard.** `wp-admin/options.php:367`:
```php
// If no settings errors were registered add a general 'updated' message.
if ( ! count( get_settings_errors() ) ) {
    add_settings_error( 'general', 'settings_updated', __( 'Settings saved.' ), 'success' );
}
```
My version omitted it, which is a straight divergence from the thing the PR claims to mirror.

**2. Other errors really can already be queued.** The handler runs on `admin_post_update` (`class-imagify-settings.php:69`), so `admin_init` has already fired and anything another plugin queued during the same request is sitting in `$wp_settings_errors`.

**Reproduced it in a test before fixing**, so this isn't taken on faith either. New integration test `testShouldNotAddSuccessNoticeWhenAnotherErrorIsAlreadyQueued` queues an error from a stand-in plugin, runs the handler, then asserts the other plugin's error survives and `settings_updated` is not queued:

```
Before the fix: FAILED - Failed asserting that an array does not contain 'settings_updated'.
After the fix:  OK (7 tests, 21 assertions)
```

Also worth recording: **the unit test was actively hiding this.** Its stub made `get_settings_errors()` return a non-empty array on every call, which is not what happens — so the guard would have looked wrong there even if I had written it correctly. It now models the real sequence: empty when the guard asks, then carrying the notice when it's read back for the transient.

Scope, for QA: only reachable on network-activated multisite. Single site is unaffected (core's own guard covers that path), and before this PR nothing rendered at all — which is the bug the PR fixes. So the false success was introduced here, exactly as the report says.

Full suites: 7 integration / 21 assertions on this handler, 528 unit, PHPCS and PHPStan clean.

PR #1258, opened 2026-08-25
Closes #1159: cover both sides of the imagify_after_optimize hook contract

### LINKED ISSUE #1159: Add integration test for imagify_after_optimize hook

Miraeld:
# Description

Fixes #1159

Adds an integration test for the `imagify_after_optimize` hook, covering **both** halves of its contract: that `MediaOptimization::task_after()` fires it, and that the Tracking `Subscriber` receives it.

No user-facing change. This protects the hook that third-party integrations (WP Rocket CDN, Amazon S3, Bulk, NextGen stats) and our own Tracking rely on.

Supersedes #1197 by @faisalahammad, which diagnosed the ticket and provided the listener-side test this builds on. Opened as a separate PR because the fix required merging 63 commits of `develop` and dropping unrelated changes, which would have meant force-pushing over a contributor's fork branch.

## Type of change

- [x] Chore
- [ ] Release

### What was tested

Three integration tests in `Test_OptimizeHookIntegration.php`:

1. **Listener side** — firing `imagify_after_optimize` reaches `Subscriber::track_media_optimized()` with both arguments intact, via the real DI-resolved Subscriber with a mocked `Tracking`.
2. **Producer side** — `MediaOptimization::task_after()` fires `imagify_after_optimize` exactly once, passing the process and the item unmodified.
3. **Producer side, dynamic hook** — `task_after()` also fires `imagify_after_{hook_suffix}` when `data.hook_suffix` is set.

**Verified by mutation testing**, which is the point of this PR:

| Mutation | Before this PR | After |
|---|---|---|
| Delete `do_action( 'imagify_after_optimize', ... )` | 436 tests green | test 2 fails |
| Delete `do_action( "imagify_after_{$hook_suffix}", ... )` | 436 tests green | test 3 fails |

A listener-only test passes even when the production call site is deleted, so it guards nothing. That was the gap.

Full suites local: **528 unit / 1528 assertions**, **148 integration / 391 assertions**, 0 failures. PHPCS and PHPStan clean.

### How to test

```bash
composer test-integration -- --filter OptimizeHookIntegration
```

To confirm the tests actually guard the hook, delete line 360 of `classes/Job/MediaOptimization.php` (`do_action( 'imagify_after_optimize', ... )`) and re-run: `testTaskAfterFiresImagifyAfterOptimizeWithProcessAndItem` must fail. Restore it afterwards.

- Test-only. No production code is modified, so no runtime behaviour changes.
- The hook under test feeds Tracking/Mixpanel, Bulk optimization status, the Next-Gen stats cache, and the Amazon S3 and WP Rocket integrations.

### Documentation

`task_after()` is `private`, and reaching it through the public `task()` requires `validate_item()` to succeed — which needs a real attachment plus a resolvable `process_class`. That is far more fixture surface than the contract under test needs, and the extra setup would itself become a source of false failures.

So the process is injected into the `optimization_process` property and `task_after()` invoked by reflection. This keeps the test on the real production code path with the smallest possible fixture, which is exactly what makes the mutation probe meaningful: the assertion fails only when the production `do_action()` changes.

### New dependencies

None.

### Risks

- Reflection couples the test to two private/protected names (`task_after`, `optimization_process`). If either is renamed the test errors loudly rather than silently passing, which is the correct failure direction.
- The process is mocked, so listeners registered by other subscribers run against a mock during the test. A media mock is supplied for that reason (sibling listeners dereference the media), matching what the original test already did.

# Mandatory Checklist

## Code validation

## Code style

N/A.

# Additional Checks

*No error handling added: the tests make no HTTP or filesystem calls, and reflection failures should surface as test errors rather than being caught.*


Miraeld:
## QA report - PASS

Ran against the local WP integration suite (WP test library + Local MySQL), plus the full unit suite.

**Acceptance criterion for #1159** - the test must fail if the hook regresses. Verified by mutation, not by inspection:

| Mutation applied to `classes/Job/MediaOptimization.php` | Result |
|---|---|
| Delete `do_action( 'imagify_after_optimize', ... )` (line 360) | `testTaskAfterFiresImagifyAfterOptimizeWithProcessAndItem` **fails** |
| Delete `do_action( "imagify_after_{$hook_suffix}", ... )` (line 378) | `testTaskAfterFiresTheSuffixedHookWhenHookSuffixIsSet` **fails** |
| No mutation | 3 tests / 12 assertions pass |

Both mutations were reverted and the tree confirmed clean afterwards.

For contrast, the same two mutations against the previous listener-only test left **all 436 unit + integration tests green** - which is why this PR exists.

**Suites**
- Integration: 148 tests, 391 assertions, 0 failures
- Unit: 528 tests, 1528 assertions, 0 failures (2 pre-existing risky - a `version_compare(null)` deprecation at `class-imagify-settings.php:374`, unrelated)
- PHPCS clean, PHPStan `[OK] No errors`
- CI: 13/13 green, including the `task-check` that was failing on #1197

**Scope check** - diff against `develop` is exactly one file, the test. No production code touched, so there is no runtime regression surface.

Credit to @faisalahammad for the original diagnosis and the listener-side test in #1197.

PR #1075, opened 2026-06-08
fix: encode email in signup AJAX and add missing Container import

faisalahammad:
## Description

Fixes signup failing for emails with special characters (`+`, `.`, `_`) by URL-encoding the email value in the signup AJAX request. Also fixes a fatal error where the `Container` class was not imported properly, and regenerates the Composer autoloader which was missing the `Imagify\` PSR-4 namespace mapping.

Fixes #1065

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue).

### What was tested

Signup flow with plus-addressed email (e.g. `hanna+may21@wp-media.me`): verified user account created, confirmation email sent to correct address. Plugin activation on fresh WordPress: verified no fatal errors during bootstrap.

### How to test

1. Open WP Admin > Settings > Imagify
2. Click "Sign up, it's free!"
3. Enter `hanna+may21@wp-media.me`
4. Submit — verify user created and email sent correctly
5. Deactivate and reactivate plugin on a fresh WP install — no `Class "Container" not found` or `Class "Imagify\Plugin" not found` errors

### Affected Features & Quality Assurance Scope

Signup flow (JS email encoding), plugin bootstrap (Container import & autoloader), bulk optimization page (PHPStan fix for filter_input). No existing optimization features are affected.

### Documentation

The signup AJAX request URL-encodes the email with `encodeURIComponent()` so that `+` characters are transmitted as `%2B` instead of being decoded as spaces by PHP. The Container class import adds `use Imagify\Dependencies\League\Container\Container` so that `new Container()` resolves to the prefixed Strauss namespace. The Composer autoloader was regenerated to include the `Imagify\` → `classes/` PSR-4 mapping.

### New dependencies

None.

### Risks

Low. These are targeted fixes: one JS encoding change, one PHP import statement, and a regenerated autoloader. The FILTER_REQUIRE_ARRAY fix in Bulk.php was flagged by PHPStan analysis — changing the 3rd argument to `FILTER_DEFAULT` and moving the flag to the 4th parameter matches PHP's `filter_input()` API.

# Mandatory Checklist

## Code validation

- [x] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.
- [x] I triggered all changed lines of code at least once without new errors/warnings/notices.
- [ ] I implemented built-in tests to cover the new/changed code.

## Code style

- [x] I wrote a self-explanatory code about what it does.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unnecessary complexity.
[...]

PR #1259, opened 2026-08-25
Fix #1065: stop mangling "+" aliases in the signup email address

### LINKED ISSUE #1065: Signup fails or emails not sent for certain email formats ("+", ".", "_")
**Describe the bug**
Signup behavior is inconsistent depending on the email format used. Certain valid email formats either fail silently or behave differently compared to signup via the plugin's website.

**To Reproduce**
1. Install and activate the Imagify plugin.
2. Click on the "Sign up, it's free!" banner.

Case 1: Emails with dots or underscore
Example: 
- film.simleu_21@gmail.com, 
- beatrix.arc.21@gmail.com 
=> user is created, but emails are not sent 

Case 2: Emails with "+" aliasing
Example:
- hanna+may21@wp-media.me
=> "Congratulations" message is shown in UI, but user is not created, no emails are sent

Control case (works as expected)
Example:
- tzinkeh@yahoo.no
=> user is created and emails are sent

Comparison with website (https://app.imagify.io/account/sign_up/)
Note: When performing the same signup cases via the website, users are created in all cases and emails are sent in all cases. 

**Expected behavior**
All valid email formats (including "+", ".", "_") should be accepted
User creation and email sending should be consistent with website behavior

Miraeld:
# Description

Fixes #1065

Signing up with a `+` alias (`hanna+may21@wp-media.me`) showed "Congratulations!" but created no account and sent no email. It now works.

What was really happening is worse than "signup fails": an account **was** created, just for a different address than the one typed, so the API key email went to a mailbox the user cannot read.

Supersedes #1075 by @faisalahammad, who found the `encodeURIComponent` fix. Opened separately because that PR's branch is on a fork with `maintainerCanModify: false`, so it could not be updated.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated - 5 new integration tests (`Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php`). They intercept `pre_http_request` and assert the address that actually leaves the plugin, which is the thing the bug corrupted:

| Case | Expectation |
|---|---|
| `hanna+may21@wp-media.me` | reaches the API unchanged |
| `film.simleu_21@gmail.com` | reaches the API unchanged |
| `hanna may21@wp-media.me` (a mangled address) | rejected, **no** account request sent |
| `  tzinkeh@yahoo.no  ` | trimmed and accepted, not rejected |
| `not-an-email` | rejected |

Verified by mutation: removing the `$email !== $raw_email` guard makes the mangled-address test fail.

Mechanism verified end to end outside the test suite too:

```
OLD  'email=' + raw            -> $_GET[email] = [hanna may21@wp-media.me] -> sanitize_email -> hannamay21@wp-media.me
NEW  'email=' + encodeURI...   -> $_GET[email] = [hanna+may21@wp-media.me] -> sanitize_email -> hanna+may21@wp-media.me
```

Suites: **153 integration / 401 assertions**, **528 unit / 1528 assertions**, 0 failures. PHPCS, PHPStan and ESLint clean. `notices.min.js` rebuilt with `grunt uglify:all` (only `notices.min.js` changed).

**Not covered:** the browser leg was not exercised manually - the local environment was not logged in and I do not enter credentials. The JS change is one expression, ESLint-clean, and confirmed present in the minified bundle; the round-trip above proves the behaviour. Worth a click-through on staging before release.

### How to test

1. Imagify settings page with no API key, click **Sign up, it's free!**.
2. Enter an address with a `+` alias, e.g. `you+imagify1@yourdomain.com`.
3. Expected: account created for **exactly** that address, and the API key email arrives at it.
4. In DevTools > Network, the `imagify_signup` request must show `email=you%2Bimagify1%40yourdomain.com` (encoded), not a bare `+`.
5. Regression: a plain address (`you@yourdomain.com`) still works, and one with stray leading/trailing spaces is accepted rather than rejected.

- Signup from the settings page and from the signup notice banner (`assets/js/notices.js`).
- `Imagify_Admin_Ajax_Post::imagify_signup_callback()` - the only consumer of the new validation.
- One new i18n string, `signupErrorRequestFailed`.
- No change to API key validation, optimization, or any other AJAX endpoint.

### Documentation

The failure had two independent causes and both are fixed.

**Client.** `'action=imagify_signup&email=' + inputValue` put the raw address in a query string. Per `application/x-www-form-urlencoded`, `+` there means a space, so PHP's `$_GET` parser decoded it to one. `encodeURIComponent()` sends `%2B` instead.

**Server.** `sanitize_email()` does not reject an address with a space - its local-part filter (`wp-includes/formatting.php:33`) *strips* characters that are invalid there. So `hanna may21@wp-media.me` silently became `hannamay21@wp-media.me`, which `is_email()` then happily accepted. That is why the UI reported success: from the plugin's point of view the signup genuinely succeeded, just for the wrong address.

Fixing only the client would leave that trap in place for any other caller, so the handler now refuses any address sanitization had to alter:

```php
$raw_email = trim( wp_unslash( $_GET['email'] ) );
$email     = sanitize_email( $raw_email );

if ( ! is_email( $email ) || $email !== $raw_email ) {
	imagify_die( __( 'Not a valid email address.', 'imagify' ) );
}
```

`sanitize_email()` does not change case, so this does not reject mixed-case addresses.

**On the dots/underscores case in the issue:** that is a separate problem. `.` and `_` are valid local-part characters, untouched by both the query-string decoding and `sanitize_email()`. The new tests prove the plugin sends such addresses through unchanged, which localises that half to the API or mail service rather than the plugin. Tracked separately so this fix is not held up by it.

### New dependencies

None.

### Risks

- **Rejecting an address that used to be accepted.** By design, but only where the old behaviour created an account under a *different* address - never a silent success turned into a silent failure, the user now gets "Not a valid email address." One edge case: a fully-qualified `user@example.com.` (trailing dot) is now rejected. Technically legal, effectively never typed, and a clear error beats a wrong account.
- **Double encoding.** Not possible: the value comes from the modal input, so it is never already percent-encoded.
- **Minified bundle drift.** Rebuilt with the project's own grunt task; the diff touches only `notices.min.js`.

# Mandatory Checklist

## Code validation

## Code style

The browser leg of the acceptance criteria was verified by the round-trip proof and the integration tests rather than by a manual click-through - see "Not covered" above.

# Additional Checks


Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
```diff
@@ -0,0 +1,169 @@
...
+ * @group  ImagifyAdminAjaxPost
+ * @group  Signup
+ */
+class Test_ImagifySignupCallback extends TestCase {
+	/**
+	 * Whether to use the Imagify API in these tests.
+	 *
+	 * @var bool
+	 */
+	protected $useApi = false; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
```
remyperona: Unused

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
```diff
@@ -0,0 +1,169 @@
...
+		add_filter(
+			'pre_http_request',
+			function ( $preempt, $args ) {
+				$this->sent_bodies[] = isset( $args['body'] ) ? $args['body'] : null;
+
+				return new \WP_Error( 'test_blocked', 'Blocked by the test suite.' );
+			},
+			10,
+			2
+		);
```
remyperona: We have a trait to mock external HTTP Requests and prevent them from happening.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
```diff
@@ -0,0 +1,169 @@
...
+		$_GET['imagifysignupnonce']     = $this->nonce;
+		$_REQUEST['imagifysignupnonce'] = $this->nonce;
+
+		try {
+			Imagify_Admin_Ajax_Post::get_instance()->imagify_signup_callback();
+		} catch ( \WPDieException $e ) {
+			return (string) $e->getMessage();
+		}
+
+		return '';
```
remyperona: The AjaxTestCase already has what is needed to simulate an AJAX request, it would be better to extend from it

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
remyperona, line 1: Test should follow one method + fixtures, as usual

remyperona (changes requested):

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
faisalahammad: Removed. The property had no effect since the base integration TestCase default was already bypassed by the rewrite.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
faisalahammad: Done. The test now uses the HttpRequestTrait from wp-media/phpunit to block and mock HTTP. We keep one small passive filter at priority 1 that only records the outgoing request body, because the trait does not expose request args and we need to assert the address reaches the API unchanged for this regression. Returning null lets every request continue to the trait's fixture-driven mock, so nothing escapes the test.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
faisalahammad: Done. The test class now extends WPMedia\PHPUnit\Integration\AjaxTestCase, sets the action to imagify_signup and calls callAjaxAction(). One extra step was needed: the plugin only registers wp_ajax_imagify_signup when wp_doing_ajax() is true at load time, so set_up() calls Imagify_Admin_Ajax_Post::get_instance()->init() now that the AJAX test case simulates that context.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
faisalahammad: Done. The five test methods are now one data driven method using the configTestData provider, and all cases moved to a fixture file at Tests/Fixtures/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php, like the other integration tests.

faisalahammad:
@Miraeld I cannot push to this branch because it lives in wp-media and my account has no write access here. The fix for all four review comments is ready as one commit on my fork: https://github.com/faisalahammad/imagify-plugin/commit/1c907a31

To apply it: `curl -sL https://github.com/faisalahammad/imagify-plugin/commit/1c907a31.patch | git am`

Or grant me write access to the branch and I will push directly. Thanks.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
Miraeld: Removed in f7c7c302. Confirmed it was doing nothing: the test passes without it, since this never touches the Imagify singleton or the api_key option, which is all `$useApi` guards.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
Miraeld: Done in f7c7c302 — now uses `HttpRequestTrait`. One passive filter remains at priority 1 that only *records* the outgoing request body and returns null, so every request still falls through to the trait's fixture-driven mock. That is needed because the trait does not expose request args, and the whole point of this regression test is asserting the address reaches the API unchanged.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
Miraeld: Done in f7c7c302 — extends `AjaxTestCase` and drives the request through `callAjaxAction()`, which also let the assertions get stronger (the response itself is now checked, not just the outgoing request). One wrinkle worth noting: the plugin only registers `wp_ajax_imagify_signup` when `wp_doing_ajax()` is true at load time, so `set_up()` calls `Imagify_Admin_Ajax_Post::get_instance()->init()` now that the AJAX context is simulated.

Tests/Integration/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php
Miraeld: Done in f7c7c302 — one `@dataProvider configTestData` method plus a fixture at `Tests/Fixtures/inc/classes/ImagifyAdminAjaxPost/imagifySignupCallback.php`, matching the other integration tests. All five cases preserved.

Miraeld:
All four review points are addressed in f7c7c302.

**Credit where it's due: @faisalahammad wrote this rework.** He replied "Done" on each thread but has no push access to this branch, so the work was sitting on his fork. I cherry-picked his commit rather than reimplementing it, so **git authorship stays with him** — `git log` shows `Author: Faisal Ahammad`.

**I reviewed and verified it rather than taking "Done" on trust.**

Honemo (approved):

PR #1056, opened 2026-05-20
fix(settings): improve clarity of picture-tag method label and warning (#1015)

Miraeld:
# Description

Fixes #1015

This change improves the clarity of the delivery-method selector in Imagify settings for next-gen image formats. The `<picture>` tag method was labelled "(preferred)", which led users — especially first-time setup users — to choose it by default without understanding the layout risks. Support consistently receives tickets about broken layouts, missing images in sliders, WooCommerce pages, and galleries caused by this method.

Two targeted copy changes are made to `views/part-settings-webp.php`:

1. **Label**: "(preferred)" → "(CDN-compatible)". This accurately reflects the technical reason to choose this option (rewrite rules do not work with CDN) without implying it is universally better.
2. **Warning message**: The existing brief note ("some themes may break") is replaced with an explicit warning that names the types of content typically affected (themes, sliders, WooCommerce pages, galleries, background images) and recommends rewrite rules for non-CDN sites.

## Type of change

- [ ] New feature (non-breaking change which adds functionality).
- [ ] Bug fix (non-breaking change which fixes an issue).
- [x] Enhancement (non-breaking change which improves an existing functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as before).
- [ ] Sub-task of #1015
- [ ] Chore
- [ ] Release

### What was tested

**Strategy C (Analysis)** — wp-env was not reachable on port 8888; Strategies A (API) and B (Browser) were skipped.

| Acceptance Criterion | Validation Method | Result | Evidence |
|---|---|---|---|
| Warning message is clearer and more explicit about possible layout impact | Code analysis — diff review of `views/part-settings-webp.php` | ✅ PASS | New text explicitly names themes, sliders, WooCommerce pages, galleries, and notes that background images are not covered. Risk is communicated before the user enables the option. |
| "preferred" label reconsidered | Code analysis | ✅ PASS | Label changed from `(preferred)` to `(CDN-compatible)`. Accurately reflects the technical reason (rewrite rules do not work with CDN) without implying universal superiority. |
| Linking to relevant documentation | Code analysis | ⚠️ PARTIAL | No confirmed docs URL exists in the codebase for this specific feature page. Non-blocking; can be added when a page is published. |

**Code quality:**
- PHPCS: zero violations
- printf placeholder count preserved (4 placeholders, all arguments unchanged)
- `esc_html__()` wrapping pattern follows existing convention in this file
- No functional logic, hooks, option names, or AJAX handlers changed

[...]

PR #1262, opened 2026-08-25
Closes #1015: stop calling <picture> tags "preferred" and spell out the layout risk

### LINKED ISSUE #1015: Enhancement: Improve Clarity Around <picture> Tag Method and “Preferred” Label for Next-Gen Delivery
**Description**
In Imagify settings, the <picture> tag method is marked as “preferred” for loading next-gen images and is selected by default.

This method replaces  tags with <picture> tags, which can cause compatibility issues with certain themes or layouts. In practice, this sometimes leads to layout breakage, missing images, or unexpected behavior.

Image

Although there is already a warning stating that some themes may break, many users do not fully understand the impact of this choice. As a result, users often proceed with the default option and later face layout issues, especially during first-time setup.

From the support side, we regularly receive tickets related to layout or display issues caused by the <picture> tag method. This is particularly problematic for new users, who may panic when their site layout breaks and immediately request a refund.

Additionally, based on support experience, rewrite rules often provide better and more consistent coverage for next-gen delivery on sites that are not using a CDN or Cloudflare.
For example:
	•	<picture> tags may not cover background images
	•	Sliders and certain WooCommerce product pages can have issues depending on the theme

**Points to Consider**
	•	The warning message could be made clearer and more explicit about possible layout impact
	•	Linking to relevant documentation could help users make a more informed choice
	•	The “preferred” label on the <picture> tag method may need reconsideration, unless there is a strong technical reason for it

Most of the tickets are under this tag: https://secure.helpscout.net/search/?query=tag%3Aimg_broken_layout

Some of them:
https://secure.helpscout.net/conversation/3182151830/611108
https://secure.helpscout.net/conversation/3144083997/601349
https://secure.helpscout.net/conversation/3165293908/607395
https://secure.helpscout.net/conversation/3151973027/603818

Miraeld:
# Description

Closes #1015

The Next-Gen delivery settings labelled the `<picture>` tag method **"(preferred)"**, while Imagify's own documentation tells users to switch *away* from it. That is where the `img_broken_layout` tickets come from: new users take the label at face value, their layout breaks, and support then moves them to rewrite rules.

The docs are unambiguous - ["My images are broken"](https://imagify.io/documentation/my-images-are-broken/) opens with:

> **TL;DR** In Imagify settings, switch WebP display setting from "Use `<picture>` tags" to "Use rewrite rules" as long as your website doesn't use a CDN or Cloudflare.

and titles its second section **"Using Rewrite Rules (Recommended)"**.

So the issue's open question - *"the 'preferred' label may need reconsideration, unless there is a strong technical reason for it"* - is answered by our own documentation: there is no such reason, and the label contradicts it.

**Copy only. No behaviour change, no setting changes value.**

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

All three points the issue raises:

| Issue point | Change |
|---|---|
| "The `preferred` label may need reconsideration" | Removed from `<picture>`; rewrite rules now marked *(recommended for sites without a CDN)* - the documentation's exact caveat, so it stays true for CDN users |
| "The warning could be clearer about layout impact" | "some themes may break" replaced with what actually breaks: theme/slider/page-builder incompatibility, hidden images, and CSS backgrounds never being covered |
| "Linking to documentation could help" | Links to the broken-images article via a new `documentation-nextgen-delivery` target |

Also added the step users currently miss: clear every cache and check the homepage, a product page, and a page with a slider.

Automated: 2 integration tests on `imagify_get_external_url()`, guarding the new target's slug so the link cannot rot silently. The URL was verified to return **HTTP 200** (the other candidate slugs I tried 404, hence the test).

Suites: **528 unit / 1528 assertions**, **150 integration / 393 assertions**, 0 failures. PHPCS and PHPStan clean.

**Not covered:** the rendered settings page was not viewed in a browser - the local environment was not logged in and I do not enter credentials. The change is text inside existing `printf()` calls whose placeholder structure is unchanged, and translator comments were updated to match. Worth an eyeball on staging, mainly to confirm the longer paragraph still reads well in the column.

### How to test

1. Imagify settings, **Optimization** tab, enable *Display images in Next-Gen format on the site*.
2. Expected: **Use rewrite rules** is labelled *(recommended for sites without a CDN)*, and **Use &lt;picture&gt; tags** carries no "(preferred)".
3. The paragraph below explains when `<picture>` is required (CDN, or unwritable config file) and what it can break.
4. "What to do if your images look broken" links to https://imagify.io/documentation/my-images-are-broken/ and opens in a new tab.
5. Regression: both options still save correctly and next-gen delivery keeps working with each.

- Settings page only, the Next-Gen delivery block (`views/part-settings-webp.php`).
- One new entry in `imagify_get_external_url()`.
- **Three translatable strings change**, so translations for them will need refreshing.
- No change to how either delivery method works, to the saved option, or to the default.

### Documentation

The labels were not just vague, they were pointing the wrong way. `<picture>` is genuinely the *right* answer for CDN and Cloudflare users, and the only one available when the configuration file cannot be written - but it is the wrong default recommendation for everyone else, because it rewrites page markup and so depends on theme cooperation, and it cannot touch images set in CSS.

The new copy therefore frames it as a trade-off with the condition attached, rather than a ranking:

- rewrite rules - does not alter page code, does not work with a CDN;
- `<picture>` - required with a CDN, alters page code, can break layout, misses CSS backgrounds.

### New dependencies

None.

### Risks

- **Translations.** Three strings changed, so those fall back to English until retranslated. Unavoidable when the existing text is misleading.
- **Users trusting the new label.** Someone on a host with an unwritable config file who switches to rewrite rules will find it does not apply. The existing UI already surfaces the config-file path when it is writable, and the new copy names the unwritable case explicitly.
- No code path, option value, or default changed, so there is no functional regression surface.

### Open question for the PO

`display_nextgen_method` still defaults to `'picture'` (`inc/classes/class-imagify-options.php:38`). Given the documentation recommends rewrite rules for non-CDN sites, and most sites are non-CDN, **should the default flip?**

I deliberately did not change it here. It would alter behaviour for every new install, and rewrite rules need a writable configuration file and silently do not apply with a CDN - so a bad default fails quietly, which is worse than the current loud failure. That is a product call with a migration question attached (existing installs must not be touched), not a copy fix. Happy to do it in a follow-up if you want it.

# Mandatory Checklist

## Code validation

## Code style

The acceptance criteria are text on a settings page; they were verified by reading the rendered `printf()` arguments and by the documentation-URL tests rather than by a browser click-through - see "Not covered" above.

# Additional Checks

*Not applicable: the change is static translatable text plus one URL constant. No new observable behaviour and nothing that can throw.*

### DIFF
--- a/Tests/Integration/inc/functions/imagifyGetExternalUrl.php
+++ b/Tests/Integration/inc/functions/imagifyGetExternalUrl.php
@@ -14,29 +14,19 @@ use Imagify\Tests\Integration\TestCase;
  */
 class Test_ImagifyGetExternalUrl extends TestCase {
 	/**
-	 * Whether to use the Imagify API in these tests.
+	 * The documentation-nextgen-delivery target is what the settings page links to
+	 * when explaining the layout risk of the <picture> tag method, so its slug must
+	 * not drift silently.
 	 *
-	 * @var bool
-	 */
-	protected $useApi = false; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
-
-	/**
-	 * The next-gen delivery documentation target resolves to the broken-images article.
+	 * @dataProvider configTestData
 	 *
-	 * That article is what the settings page links to when explaining the layout
-	 * risk of the <picture> tag method, so the slug must not drift silently.
+	 * @param array  $config   Target and optional query args.
+	 * @param string $expected Expected URL.
 	 */
-	public function testShouldReturnTheNextGenDeliveryDocumentationUrl() {
+	public function testShouldReturnExpectedUrl( $config, $expected ) {
 		$this->assertSame(
-			IMAGIFY_SITE_DOMAIN . '/documentation/my-images-are-broken/',
-			imagify_get_external_url( 'documentation-nextgen-delivery' )
+			$expected,
+			imagify_get_external_url( $config['target'], $config['query_args'] ?? [] )
 		);
-	}
-
-	/**
-	 * An unknown target still returns an empty string.
-	 */
-	public function testShouldReturnAnEmptyStringForAnUnknownTarget() {
-		$this->assertSame( '', imagify_get_external_url( 'no-such-target' ) );
 	}
 }
--- a/Tests/Integration/inc/functions/imagifyGetExternalUrl.php
+++ b/Tests/Integration/inc/functions/imagifyGetExternalUrl.php
@@ -1,0 +1,42 @@
+<?php
+declare(strict_types=1);
+
+namespace Imagify\Tests\Integration\inc\functions;
+
+use Imagify\Tests\Integration\TestCase;
+
+/**
+ * Tests for imagify_get_external_url().
+ *
+ * @covers ::imagify_get_external_url
+ *
+ * @group  Functions
+ */
+class Test_ImagifyGetExternalUrl extends TestCase {
+	/**
+	 * Whether to use the Imagify API in these tests.
+	 *
+	 * @var bool
+	 */
+	protected $useApi = false; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
+
+	/**
+	 * The next-gen delivery documentation target resolves to the broken-images article.
+	 *
+	 * That article is what the settings page links to when explaining the layout
+	 * risk of the <picture> tag method, so the slug must not drift silently.
+	 */
+	public function testShouldReturnTheNextGenDeliveryDocumentationUrl() {
+		$this->assertSame(
+			IMAGIFY_SITE_DOMAIN . '/documentation/my-images-are-broken/',
+			imagify_get_external_url( 'documentation-nextgen-delivery' )
+		);
+	}
+
+	/**
+	 * An unknown target still returns an empty string.
+	 */
+	public function testShouldReturnAnEmptyStringForAnUnknownTarget() {
+		$this->assertSame( '', imagify_get_external_url( 'no-such-target' ) );
+	}
+}
--- a/Tests/Fixtures/inc/functions/imagifyGetExternalUrl.php
+++ b/Tests/Fixtures/inc/functions/imagifyGetExternalUrl.php
@@ -1,0 +1,48 @@
+<?php
+
+return [
+	'test_data' => [
+		'shouldReturnTheNextGenDeliveryDocumentation' => [
+			'config'   => [
+				'target' => 'documentation-nextgen-delivery',
+			],
+			'expected' => 'https://imagify.io/documentation/my-images-are-broken/',
+		],
+
+		'shouldReturnTheDocumentationHome'            => [
+			'config'   => [
+				'target' => 'documentation',
+			],
+			'expected' => 'https://imagify.io/documentation/',
+		],
+
+		'shouldReturnTheImagickGdDocumentation'       => [
+			'config'   => [
+				'target' => 'documentation-imagick-gd',
+			],
+			'expected' => 'https://imagify.io/documentation/solve-imagemagick-gd-required/',
+		],
+
+		'shouldReturnTheSubscriptionAppUrl'           => [
+			'config'   => [
+				'target' => 'subscription',
+			],
+			'expected' => 'https://app.imagify.io/#/subscription',
+		],
+
+		'shouldAppendQueryArgs'                       => [
+			'config'   => [
+				'target'     => 'documentation',
+				'query_args' => [ 'utm_source' => 'imagify' ],
+			],
+			'expected' => 'https://imagify.io/documentation/?utm_source=imagify',
+		],
+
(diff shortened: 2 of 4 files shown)


Miraeld:
**Converting to draft — this needs product/marketing sign-off on the wording before it can merge.**

I found #1056 after opening this: it did the same thing for #1015 and was closed by @remyperona for exactly the reason that applies here:

> The issue I see here is that the copywriting changes have not been clearly defined nor validated by whoever would be responsible about it in the first place (product and/or marketing). We let the AI decide based on the issue description, but what is contained in it is only examples.

That objection stands against this PR too. I wrote the replacement copy from the issue description, which is examples rather than approved wording. So treat the strings here as a **proposal**, not a deliverable.

One piece of evidence #1056 did not have, which is worth putting in front of whoever owns the copy, because it makes part of this a factual inconsistency rather than a wording preference:

[Imagify's own documentation](https://imagify.io/documentation/my-images-are-broken/) opens with

> **TL;DR** In Imagify settings, switch WebP display setting from "Use `<picture>` tags" to "Use rewrite rules" as long as your website doesn't use a CDN or Cloudflare.

and titles its second section **"Using Rewrite Rules (Recommended)"**.

So the UI's "(preferred)" on `<picture>` contradicts our published documentation. Whatever the final wording is, those two should agree — the issue asked whether there was "a strong technical reason" for the label, and the docs answer no.

Three things need a decision from product/marketing:

1. The exact label on each radio option (this PR proposes dropping "(preferred)" and marking rewrite rules *"(recommended for sites without a CDN)"*).
2. The exact warning wording for the `<picture>` method.
3. Whether to link https://imagify.io/documentation/my-images-are-broken/ from the settings page. Note #1056's QA flagged that no docs URL was linked "pending a page being published" — this page exists and returns 200, so that blocker may be gone.

Also worth noting #1056 left an e2e spec on its branch (`Tests/e2e/specs/settings-webp-method.spec.ts`) asserting the old proposed strings. If wording is approved, that spec should be revived and updated rather than rewritten.

Happy to apply whatever copy comes back — the mechanical part is done and tested.

Tests/Integration/inc/functions/imagifyGetExternalUrl.php
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @group  Functions
+ */
+class Test_ImagifyGetExternalUrl extends TestCase {
+	/**
+	 * Whether to use the Imagify API in these tests.
+	 *
+	 * @var bool
+	 */
+	protected $useApi = false; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
```
remyperona: Why is this included?

Tests/Integration/inc/functions/imagifyGetExternalUrl.php
```diff
@@ -0,0 +1,42 @@
...
+			imagify_get_external_url( 'documentation-nextgen-delivery' )
+		);
+	}
+
+	/**
+	 * An unknown target still returns an empty string.
+	 */
+	public function testShouldReturnAnEmptyStringForAnUnknownTarget() {
+		$this->assertSame( '', imagify_get_external_url( 'no-such-target' ) );
+	}
```
remyperona: Should be using one method + fixtures here as well

remyperona (changes requested):

Tests/Integration/inc/functions/imagifyGetExternalUrl.php
Miraeld: Good catch — it was cargo-culted from the surrounding test files, and it is not needed. Verified rather than assumed: I removed it and the test still passes, because this exercises a pure function that never touches the Imagify singleton or the `api_key` option, which is all `$useApi` guards.

Removed in e5a220c3.

Tests/Integration/inc/functions/imagifyGetExternalUrl.php
Miraeld: Done in e5a220c3 — single method with `@dataProvider configTestData` plus a fixture at `Tests/Fixtures/inc/functions/imagifyGetExternalUrl.php`, matching `ImagifySettings/printText` and `ImagifySettings/updateSiteOptionOnNetwork`.

Since adding a case is now one fixture entry, I widened it from 2 cases to 6: the new `documentation-nextgen-delivery` target, the documentation home, `documentation-imagick-gd`, an app-domain target (`subscription`), the query-args path, and an unknown target. That guards the whole switch rather than only the slug this PR adds.

remyperona (approved):

sandyfzu (changes requested):
All good from QA perspective, but I think this will require product validation.

For me, the text seems to be too much.

@raahulbasu, could you please validate?

I attached the built zip.

[imagify-2.3.3.zip](https://github.com/user-attachments/files/31521938/imagify-2.3.3.zip)

sandyfzu (approved):

Issue #1264, opened 2026-08-25
Nothing builds the _dev bud pipeline, so assets/admin/js/* drifts silently


Miraeld:
Found while surveying the JS build for #1215.

## The problem

`_dev/src/bulk.js` and `_dev/src/chart.js` are compiled by bud/webpack into `assets/admin/js/{bulk,chart,runtime}.js`, per `_dev/bud.config.js`.

**Nothing runs that build.** Verified:

| Candidate | Builds bud? |
|---|---|
| `npm run build` (root) | No — `grunt` default is `css` + `minify`, which only covers `assets/js/*.js` |
| `bin/build-zip.sh` | No — runs `npm run build` at the repo root only |
| `.github/workflows/*` | No — no workflow invokes bud |
| `deploy-tag.yml` | No — runs only `composer install -o --no-dev` |

`_dev/` has its own separate `package.json` (`@roots/bud ^6.11.0`) with its own `build` script, which a developer must remember to run by hand, in that directory.

## Why it matters

`assets/admin/js/*.js` is **committed** build output (real commit history, most recently via #1208). So the shipped bundle is whatever the last person to run bud by hand happened to commit.

Concretely:
- A PR can change `_dev/src/bulk.js` and not rebuild. CI passes. The source and the shipped bundle silently disagree, and the source change simply has no effect in production.
- Nothing detects it. There is no CI check that the committed bundle matches its source.
- `deploy-tag.yml` performs no JS build, so whatever is committed is what reaches WordPress.org.

This is a live footgun rather than a hypothetical: the bulk optimization page is driven entirely by that bundle.

## Suggested fix

Either:

1. **Wire it in** — have the root `npm run build` also run the bud build (`npm --prefix _dev run build`, or fold bud into the root `package.json`), and add a CI job that rebuilds and fails if `git diff --exit-code assets/admin` is dirty. That catches the drift at PR time.
2. **Or build at release** — add a JS build step to `deploy-tag.yml` and stop committing built output entirely. Note this is a bigger change than it looks: Composer installs (see #1073) also never run a JS build, so committed bundles would still be needed there unless that path is addressed too.

Option 1 is the smaller, safer change and fixes the immediate hazard.

## Related

- #1215 — JS testing in CI; this was found while mapping that work. The two overlap: consolidating the pipelines there would naturally absorb this.
- #1073 — Composer installs never run a build, which constrains option 2.

Issue #1263, opened 2026-08-25
sanitizeContext() sanitizes nothing: a string literal is passed to .replace() instead of a RegExp


Miraeld:
Found while surveying the JS for #1215.

## The bug

`assets/js/files-list.js:311` and `assets/js/media-modal.js:297` are identical:

```js
sanitizeContext: function( context ) {
    context = context.replace( '/[^a-z0-9_-]/gi', '' ).toLowerCase();
    return context ? context : 'wp';
},
```

The first argument is a **string literal**, not a RegExp. `String.prototype.replace` with a string needle does exact-substring matching, so it looks for the literal text `/[^a-z0-9_-]/gi` and finds it never. The function strips nothing — it only lowercases.

Demonstrated:

| Input | Actual | Expected |
|---|---|---|
| `WP<script>` | `wp<script>` | `wpscript` |
| `a b!c` | `a b!c` | `abc` |
| `custom-folders` | `custom-folders` | `custom-folders` |

The fix is to drop the quotes: `context.replace( /[^a-z0-9_-]/gi, '' )`.

## Severity: low, and deliberately so

**This is not exploitable.** The context is re-sanitized server-side with `sanitize_key()` — see `imagify_sanitize_context()` in `inc/functions/common.php:51`, called at `inc/classes/class-imagify-admin-ajax-post.php:1207`. So the server never trusts the client value.

What this actually is: a **dead defence-in-depth layer**. It has been silently doing nothing, and anyone reading the code would reasonably assume the client sanitizes before sending.

## Why it is worth fixing anyway

- It is two lines, in two files.
- It is exactly the class of bug #1215 exists to catch — parsing/sanitization that silently no-ops — so it makes a good first test case for the new JS suite.
- Removing the quotes changes what gets sent to the server for exotic context values. Worth a quick check that no caller relies on the current pass-through behaviour before merging.

## Acceptance criteria

- Both copies use a real RegExp.
- A unit test asserts characters outside `[a-z0-9_-]` are stripped.
- Confirmed that no code path depends on the old no-op behaviour.

## Notes

The two copies are byte-identical, as are the two `sanitizeId` implementations next to them (`files-list.js:301`, `media-modal.js:287`). Deduplicating them is a natural follow-up once #1215 gives the JS a module seam.

PR #1265, opened 2026-08-25
Closes #1215: run JavaScript unit tests in CI

### LINKED ISSUE #1215: Set up JavaScript testing and run it in CI
## Problem

The plugin ships a meaningful amount of JavaScript and **none of it is tested**. CI runs PHPUnit, PHPStan, PHPCS and Playwright, but no JavaScript test job exists at all. Any regression in the admin JS is caught only by manual QA or by users.

This surfaced while fixing #760: the bug was a two-line arithmetic error in the settings-page progress bar that had been shipping since the feature was introduced, and there was nowhere to put a regression test.

## Why it is not just "add a test runner"

There are currently **two parallel JS pipelines**, and the untested code lives in the older one:

| | Files | Build | Module system | Tests |
|---|---|---|---|---|
| Legacy | `assets/js/*.js` — `options.js`, `media-modal.js`, `library.js`, … | `Gruntfile.js`: `eslint` + `uglify` → `*.min.js` | none, jQuery IIFEs against globals | none |
| Modern | `_dev/src/bulk.js`, `_dev/src/chart.js` | `_dev/bud.config.js` (`@roots/bud`, webpack) → `assets/admin/` | ES modules | none |

The legacy files are not modules: they are IIFEs that attach to `window.imagify` and read globals such as `imagifyOptions` and `jQuery`. Nothing can `require`/`import` them, so there is no clean seam to test against.

A stopgap exists today in `Tests/js/getProgress.test.js` (added in #1214): it loads the real `assets/js/options.js` inside a `node:vm` context behind hand-written stubs, including a Proxy-based chainable jQuery fake. It works and it caught the #760 regression under mutation testing, but it is **deliberately not wired into CI**, because:

- the stub only satisfies the current load path — an unrelated edit to `options.js` that reads `.length` to drive a branch, uses `instanceof`, or touches an unstubbed global breaks the *harness*, producing a confusing failure in a file the author never touched;
- it boots the whole bundle to reach one pure function;
- it does not generalise — anything DOM-touching needs jsdom regardless;
- VM objects cross realms, so `assert.deepStrictEqual` rejects structurally identical values and every assertion needs unwrapping.

Gating CI on that would be a band-aid. The seam needs fixing first.

## Proposed approach

1. **Unify the pipelines.** Migrate the remaining `assets/js/*.js` sources into `_dev/src/`, built by bud into `assets/`. One build, one lint config, real modules. This is the bulk of the work and can be done file by file.
2. **Add a test runner.** Vitest is the natural fit for a webpack/bud project — jsdom available when a test needs the DOM, plain node environment when it does not, coverage built in. Jest is an acceptable alternative if preferred.
3. **Add one CI job** running the JS suite, alongside the existing PHP jobs. One job covering all JS, not a per-file bolt-on.
4. **Port the stopgap.** Once `options.js` is a module, rewrite `Tests/js/getProgress.test.js` as a plain import and delete the VM/Proxy loader.
5. **Bring test sources under lint.** The `eslint` target in `Gruntfile.js` currently covers `gruntfile.js` and `assets/js/*.js` only, so test files are unlinted.

## Suggested slicing

This does not need to land at once:

- **Slice 1** — add the runner, the CI job, and tests for the already-modular `_dev/src/bulk.js` and `_dev/src/chart.js`. Delivers a working gate immediately without touching the legacy files.
- **Slice 2** — migrate `options.js` to `_dev/src/`, port the #1214 stopgap to a normal import, drop the VM loader.
- **Slice 3** — migrate the remaining legacy `assets/js/*.js` files.

## Acceptance criteria

- A JS test suite runs in CI on every PR and fails the build when a test fails.
- Adding a test for a piece of admin JS does not require bespoke loaders, VM contexts, or hand-written global stubs.
- `_dev/src/bulk.js` and `_dev/src/chart.js` have at least smoke-level coverage.
- The stopgap `Tests/js/getProgress.test.js` is replaced by a straightforward module test, and the `node:vm` loader is removed.
- Test sources are linted by the same config as production sources.
- The build output (`assets/…`) is unchanged by the migration, verified by diffing the built files before and after.

## Notes

- `package.json` currently exposes `npm run test:js`, which runs only the stopgap via Node's built-in runner. It should be replaced, not extended.
- Worth deciding as part of this whether minified assets stay committed to the repository, since the migration touches how they are produced.

## Related

- #760 / #1214 — the bug that exposed the gap, and the stopgap test

Miraeld:
# Description

Refs #1215 — **slice 1 of 3. This does not close the ticket.**

CI had no JavaScript job at all. This adds one, plus the first tests that actually guard something.

No user-facing change. One production refactor (pure functions moved into a new file) with the `window.imagify` contract preserved.

## Type of change

- [x] Sub-task of #1215
- [ ] Chore
- [ ] Release

### What was tested

**18 JS tests, 2 suites, passing.** And they have teeth — three mutations of the extracted arithmetic were each caught:

| Mutation | Result |
|---|---|
| `quota >= 1000` → `quota > 1000` (GB threshold) | **CAUGHT** |
| `(100 - couponValue)` → `(100 + couponValue)` (discount sign) | **CAUGHT** |
| Removed the 2-decimal rounding on the monthly-from-annual price | **CAUGHT** |

Also verified by simulating production load order that `window.imagify.humanSize` still resolves after the move, since `bulk.js:454` depends on it.

Unchanged elsewhere: **528 PHP unit / 1528 assertions**, **148 integration / 391 assertions**. ESLint clean. Minified assets rebuilt with the project's own grunt task — only the three touched files changed.

### How to test

```bash
nvm use            # 22, from .nvmrc
npm ci
npm run test:unit  # 18 tests, 2 suites
```

To confirm the tests actually guard the maths, flip `quota >= 1000` to `quota > 1000` in `assets/js/helpers.js` and re-run — `formatQuota` must fail. Restore afterwards.

Manual regression check, since the pricing modal is revenue-facing: open the pricing modal, confirm plan quotas still read "1 GB"/"500 MB"/"Unlimited", and that a percentage promo still shows the discounted monthly and yearly prices. Then check a bulk optimization run still shows human-readable sizes (that path uses the moved `humanSize`).

- **Pricing modal** — quota label and discounted price display now come from `helpers.js`. Highest-risk area in this PR; please eyeball it.
- **Bulk optimization** — consumes `window.imagify.humanSize`, which moved file but not location on `window.imagify`.
- **CI** — one new job. `e2e.yml` now takes its Node version from `package.json` instead of a hardcoded `22`; same version, one source.
- No PHP behaviour change. `class-imagify-assets.php` only registers one more script.

### Documentation

**Tooling follows remyperona's comment on the ticket, not the ticket body.** He specified Jest via `@wordpress/scripts`; the body proposed Vitest with bud. Three non-obvious choices came out of that:

- **`@wordpress/scripts` pinned to `^31`.** Version 34 requires `eslint ^10`, which cannot coexist with `grunt-eslint@24`'s eslint 8. `^31` wants `eslint ^8.3` and dedupes onto the 8.27 already installed. Taking latest would have broken `grunt eslint`.
- **Node 22, pinned once.** `engines` + `.nvmrc`, and `e2e.yml` switched to `node-version-file`. 22 rather than BackWPup Pro's 20 because wp-scripts only needs `>=18.12` and `Tests/e2e/package.json` already declares `@types/node ^22` — pinning 20 would have silently downgraded a passing Playwright job.
- **No `--passWithNoTests`.** BackWPup Pro carries that flag and has zero test files, so its Jest job is green while testing nothing. That is precisely the failure mode this ticket exists to fix.
- **Jest is explicitly scoped** by `Tests/js/jest.config.js`. A default `testMatch` also collects the 12 Playwright specs in `Tests/e2e/specs/*.spec.ts`, which import `@playwright/test` and fail under Jest.

**Why a new file rather than tests against the existing ones.** Every legacy file is a jQuery IIFE with no export, so nothing can be imported. `assets/js/options.js` is only reachable today through a `node:vm` sandbox with a Proxy-based fake jQuery — the ticket rightly calls that a band-aid. Rather than add more of it, `assets/js/helpers.js` gives the JS its first real seam: pure functions, no jQuery, no DOM, no `imagify*` globals, so a test just calls `require()`.

It holds `humanSize` (moved out of `admin.js`, still assigned to `window.imagify` so `bulk.js` is unaffected) plus `formatQuota`, `monthlyFromAnnual` and `applyDiscount`, extracted verbatim from `pricing-modal.js`. That arithmetic decides the prices customers see and had no tests whatsoever.

`Tests/js/getProgress.test.js` is ported from `node:test` to Jest so the #760/#865 regression coverage runs in CI rather than behind a separate runner. Its `node:vm` loader survives untouched — it dies in slice 2, per the ticket.

### New dependencies

`@wordpress/scripts ^31.2.0` (devDependency), matching backwpup-pro.

### Risks

- **Pricing display.** The extracted maths is a verbatim move — `applyDiscount` reproduces `(100 - value)/100` and the same rounding — and is mutation-tested. Still the thing to check by hand before merge.
- **Script load order.** `helpers` is registered with no dependencies and `admin` now depends on it, so it always loads first. `pricing-modal` depends on `admin`, so it inherits it transitively.
- **`window.imagify.humanSize` disappearing.** It does not: `helpers.js` assigns it, and the load order above guarantees it exists before `admin.js` or `bulk.js` run. Verified by simulation.
- **A third toolchain.** wp-scripts is test-only here; grunt and bud keep their current jobs untouched. Consolidation is a later slice.

### What this does NOT do

Being explicit, because the ticket's slice 1 promised more than is achievable:

- `_dev/src/bulk.js` and `_dev/src/chart.js` have **no coverage**. `bulk.js` is not modular — zero `import`/`export`, reads `imagifyBulk` at module-evaluation time (~line 89), calls `init()` unconditionally at line 1227, and can fire a `swal` on import. Testing it today needs exactly the jsdom + jQuery + global-stub harness the ticket wants abolished, so its coverage genuinely depends on the slice-2 migration.
- The `node:vm` loader still exists.
- `_dev/src` has not moved to `src/`. Deferred to slice 2 so the bud config, `.distignore` and the modularisation move together — note `.distignore` is a denylist, so a new top-level `src/` would be shipped to WordPress.org unless added to it.

### Found on the way, filed separately

- #1263 — `sanitizeContext()` passes a string literal to `.replace()` instead of a RegExp, so it strips nothing. Not exploitable (the server re-sanitizes) but a dead defence-in-depth layer.
- #1264 — nothing builds the `_dev` bud pipeline, so `assets/admin/js/*` drifts silently from its source.

# Mandatory Checklist

## Code validation

## Code style

The acceptance criteria met are slice 1's; #1215's remaining criteria are listed under "What this does NOT do" and stay open.

# Additional Checks

*Not applicable: the new code is pure arithmetic and string formatting with no I/O.*


Miraeld:
## bulk.js is now covered — 44 tests

You asked whether testing `bulk.js` was possible. It is, and it turned out to be the *cheapest* file to cover rather than the hardest, for one reason: **`bulk.js` is already a webpack entry via bud**, so it can `import` a module with no change to the build architecture. The nine legacy `assets/js/*.js` files cannot — that is what makes them expensive, not their size.

**Safety check first.** Nothing automates the bud build (#1264), so `assets/admin/js/bulk.js` is a hand-built committed artifact. Before touching the source I installed bud and rebuilt the *unmodified* file: the output was **byte-identical** to what is committed. So the bundle diff in this PR contains only these changes, and is reviewable.

**Extracted into `_dev/src/lib/bulk-state.js`:**

| Function | Was | Why it matters |
|---|---|---|
| `findBlockingError()` | 6 near-identical branches in `hasBlockingError` | The **priority order** decides which error the user sees when several apply. And `! keyIsValid` is the sharp edge — a missing flag must block; inverting it would let a run start unauthenticated |
| `buildAjaxUrl()` | inline in `getAjaxUrl` | Guards `Number.isInteger( item.level )`. **Level 0 is a real optimization level**; a truthiness check would silently drop it and change what gets optimized |
| `normalizeChartData()` | duplicated in two chart paths | All-zero doughnut guard. Now shared, and no longer mutates its input |

`hasBlockingError()` keeps only the code-to-modal mapping, which genuinely needs the DOM.

**Mutation-verified** — each of these fails the suite:
- inverting the `keyIsValid` negation
- replacing `Number.isInteger` with a truthiness test
- moving the forced chart slice from index 0 to 1
- renaming the highest-priority blocking condition

**Also:** the new sources are now linted (grunt eslint covers `_dev/src/lib` and `Tests/js`, with an `.eslintrc` override for ES modules and jest globals), and `_dev/package-lock.json` is committed for the first time so the bundle build is reproducible from a lockfile.

### Two things I deliberately left

**The pre-existing `_dev/src/*.js` files are still unlinted.** Adding them to the eslint target produced **848 style violations** — they have never been linted. Fixing those here would bury this diff and risk perturbing the bundle. Worth its own cleanup PR.

**The nine legacy `assets/js/*.js` files are still IIFEs.** Their pure logic is extracted and tested, but the DOM/jQuery-coupled bulk of them is not, and converting them to modules would rewrite every emitted asset with no test able to prove equivalence. Still my recommendation to do that separately, if at all.

Running total: **44 JS tests**, up from 0 at the start of this PR.

remyperona (approved):

